PR #1, opened 2015-03-09
create substitution accumulo graph

rygim:
Also some refactoring inside of the accumulo graph to allow it to be extended


joeferner:
Is there a way we can pull this off without extending AccumuloGraph and always having a NamingStrategy on the graph that always gets called. Then have a default implementation that is an IdentityNamingStrategy that returns the variable passed in.

rygim:
Joe, I pushed a change to this, please check the tests to make sure that they work, I was having trouble with mini accumulo earlier but I can't replicate it currently.  Let me know if there is anything else you would like me to do.

joeferner:
I just moved some things around. Mainly the substitution stuff into core so that we can eventually if we want to use it for ElasticSearch as well.

PR #39, opened 2015-09-29
support in process elasticsearch node

joeferner:


elasticsearch-base/src/main/java/org/vertexium/elasticsearch/ElasticSearchSearchIndexBase.java
```diff
@@ -50,19 +55,76 @@
...
+                                .put("gateway.type", "local")
+                                .put("index.number_of_shards", numberOfShards)
+                                .put("index.number_of_replicas", "0")
+                                .put("path.data", dataPath)
+                                .put("path.logs", logsPath)
+                                .put("path.work", workPath)
+                ).node();
+        inProcessNode.start();
+        Client client = inProcessNode.client();
+        while (true) {
```
srfarley: How about a maximum number of retries instead of looping forever?

joeferner: done

PR #40, opened 2015-09-30
findRelatedEdgeSummary was returning soft deleted and hidden edges, s…

sfeng88:
…o now we are filtering them out


accumulo-iterators/src/main/java/org/vertexium/accumulo/iterator/EdgeRefFilter.java
```diff
@@ -1,21 +1,29 @@
 package org.vertexium.accumulo.iterator;
 
+import com.google.common.collect.Lists;
```
joeferner: we should not rely on guava here. Guava is transitively pulled in from Accumulo and I don't want to rely on them, I would like to keep the iterator as standalone as possible.

PR #42, opened 2015-10-02
Changed ValueSerializer to VertexiumSerializer to be more generic

joeferner:
Changed `ValueSerializer` to `VertexiumSerializer`. This requires a change to your configuration:
  `accumulo.graph.valueSerializer` is now `accumulo.graph.serializer` and the class name specified should change as well.


kunklejr:
What's the purpose of the change? I imagine it's more than just a name change.

joeferner:
I would like the future JDBC backend to also use it and currently the serializers are tied to Accumulo via the `Value` class

kunklejr:
I see. Thanks for the context. That makes sense.

srfarley:
Looks good!

PR #46, opened 2015-10-06
SQL/JDBC implementation

srfarley:
@joeferner: I intend to squash the commits before merge.

### DIFF
--- a/sql/src/main/java/org/vertexium/sql/SqlGraphConfiguration.java
+++ b/sql/src/main/java/org/vertexium/sql/SqlGraphConfiguration.java
@@ -1,0 +1,102 @@
+package org.vertexium.sql;
+
+import com.google.common.base.Preconditions;
+import org.apache.commons.dbcp2.BasicDataSource;
+import org.vertexium.VertexiumSerializer;
+import org.vertexium.inmemory.InMemoryEdge;
+import org.vertexium.inmemory.InMemoryGraphConfiguration;
+import org.vertexium.inmemory.InMemoryTableElement;
+import org.vertexium.inmemory.InMemoryVertex;
+import org.vertexium.sql.collections.SqlMap;
+
+import javax.sql.DataSource;
+import java.util.HashMap;
+import java.util.Map;
+
+public class SqlGraphConfiguration extends InMemoryGraphConfiguration {
+    protected static final String CONFIG_JDBC_DRIVER_CLASS = "sql.jdbc.driverClass";
+    protected static final String CONFIG_JDBC_URL = "sql.jdbc.url";
+    protected static final String CONFIG_JDBC_USERNAME = "sql.jdbc.username";
+    protected static final String CONFIG_JDBC_PASSWORD = "sql.jdbc.password";
+    protected static final String KEY_COLUMN_NAME = "id";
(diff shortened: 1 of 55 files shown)


sql/src/main/java/org/vertexium/sql/SqlGraphConfiguration.java
```diff
@@ -0,0 +1,137 @@
...
+        dataSource.setUrl(url);
+        dataSource.setUsername(username);
+        dataSource.setPassword(password);
+        Handle handle = new DBI(dataSource).open();
+        vertexMap = newVertexMap(handle);
+        edgeMap = newEdgeMap(handle);
+        metadataMap = newMetadataMap(handle);
+    }
+
+    private VertexiumSerializer createSerializer(Map<String, Object> config) {
```
joeferner: I know I put this here but should we move this to GraphConfiguration similar to the table name moving there?

sql/src/main/java/org/vertexium/sql/SqlGraphConfiguration.java
```diff
@@ -0,0 +1,137 @@
...
+        String driverClass = getConfigString(config, CONFIG_JDBC_DRIVER_CLASS);
+        String url = getConfigString(config, CONFIG_JDBC_URL);
+        String username = getConfigString(config, CONFIG_JDBC_USERNAME);
+        String password = getConfigString(config, CONFIG_JDBC_PASSWORD);
+        dataSource = new BasicDataSource();
+        dataSource.setDriverClassName(driverClass);
+        dataSource.setUrl(url);
+        dataSource.setUsername(username);
+        dataSource.setPassword(password);
+        Handle handle = new DBI(dataSource).open();
```
joeferner: I find it a little strange to open the Handle here? Should this be moved to SqlGraph?

joeferner:
Can you also update the CHANGELOG.md.  Other than that it looks good.

sql/src/main/java/org/vertexium/sql/SqlGraphConfiguration.java
srfarley: I didn't want to have a dependency on the JDBI library anywhere else but here. But having reviewed this more, I think I need to create a new `Handle` instance within every call in `SqlMap` (wrapped by an auto-closing try block). That means I'll need to provide `SqlMap` with a `DBI` instance.

sql/src/main/java/org/vertexium/sql/SqlGraphConfiguration.java
srfarley: done

sql/src/main/java/org/vertexium/sql/SqlGraphConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+
+    protected SqlMap<Object> metadataMap() {
+        return metadataMap;
+    }
+
+    protected String tableNameWithPrefix(String tableName) {
+        return getTableNamePrefix() + "_" + tableName;
+    }
+
+    private SqlMap<InMemoryTableElement<InMemoryEdge>> newEdgeMap() {
```
joeferner: This still seems like an odd place for these methods. Shouldn't these be in SqlGraph?

PR #48, opened 2015-10-13
add IP address type

joeferner:


core/src/main/java/org/vertexium/type/IPAddress.java
```diff
@@ -0,0 +1,81 @@
+package org.vertexium.type;
+
+import org.vertexium.VertexiumException;
+
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class IPAddress implements Serializable, Comparable<IPAddress> {
```
srfarley: I recommend calling this `IpV4Address` for clarity, and to be able to later add `IpV6Address`.

PR #49, opened 2015-10-24
sql streaming properties are now stored in a separate db table

srfarley:


sql/src/main/java/org/vertexium/sql/SqlStreamingPropertyValue.java
```diff
@@ -0,0 +1,63 @@
...
+    public InputStream getInputStream() {
+        try (Handle handle = dbi.open()) {
+            Row row = handle
+                    .createQuery(String.format(
+                            "select %s from %s where %s = ?",
+                            VALUE_COLUMN_NAME, tableName, KEY_COLUMN_NAME))
+                    .bind(0, makeId(elementId, key, name, visibility, timestamp))
+                    .map(new RowResultSetMapper()).first();
+
+            return new AutoDeleteFileInputStream(row.inputStream);
```
joeferner: I'm a little confused here why do we need to copy it to a file. I'm guessing the input stream gets closed when the Handle closes? For a future improvement we might want to get fancy here and read the first n bytes and if it's below some threshold just store it in memory (ByteArrayInputStream), if it's above that threshold write it to disk.

srfarley: You're correct - the input stream should not be used after the database connection is closed. I'll implement your suggestion. What's a reasonable threshold? 100k? 1000k?

joeferner: 1meg seems reasonable. That'll catch most text documents.

PR #52, opened 2015-12-14
add serialVersionUID to SQL serializable classes

joeferner:


sfeng88:
LGTM

joeferner:
I think Travis screwed up. All test run fine.

Looking at the logs it looks like the process was "killed"

sql/src/main/java/org/vertexium/sql/SqlTableEdge.java
```diff
@@ -39,6 +41,7 @@ InMemoryTableEdge asInMemoryTableElement() {
 
     private static class InMemoryAdapter extends InMemoryTableEdge
             implements Storable<SqlTableElement<InMemoryEdge>, SqlGraph> {
+        private static final long serialVersionUID = SqlTableEdge.serialVersionUID;
```
srfarley: What's the reason for this `serialVersionUID` to be the same as its outer class?

joeferner: Does this class ever get serialized?  Is it possible to throw an exception instead when it tries to get serialized? 

On December 14, 2015 6:51:34 PM EST, Steven Farley <notifications@github.com> wrote:
>> @@ -39,6 +41,7 @@ InMemoryTableEdge asInMemoryTableElement() {
>>  
>>      private static class InMemoryAdapter extends InMemoryTableEdge
>>              implements Storable<SqlTableElement<InMemoryEdge>,
>SqlGraph> {
>> +        private static final long serialVersionUID =
>SqlTableEdge.serialVersionUID;
>
>What's the reason for this `serialVersionUID` to be the same as its
>outer class?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/v5analytics/vertexium/pull/52/files#r47578126

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

joeferner:
@srfarley I now throw an exception in the adapters if they try to be serialized.

srfarley:
I'm not thrilled about adding this because it does nothing for serializers that don't use Java serialization, like our `XStreamVertexiumSerializer`. I'm pretty sure these static inner classes are never serialized.

joeferner:
I'm just bothered by the fact that it's marked as Serializable but if it gets serialized it won't work properly.

PR #54, opened 2015-12-15
Kryo support no def constructors

srfarley:


kryo-serializer/src/main/java/org/vertexium/serializer/kryo/KryoFactory.java
```diff
@@ -27,6 +28,8 @@ public boolean useReferences(Class type) {
         registerClasses(kryo);
 
         kryo.setAutoReset(true);
+        kryo.setInstantiatorStrategy(new Kryo.DefaultInstantiatorStrategy(new StdInstantiatorStrategy()));
```
joeferner: Is this not the default?

srfarley: The default is `Kryo.DefaultInstantiatorStrategy` only, which doesn't support the absence of the default no-arg constructor. The `StdInstantiatorStrategy` is the fallback, which supports classes without the constructor by using the native JDK instantiation (also used by Java serialization).

joeferner:
LGTM

PR #55, opened 2016-01-06
fix in memory getEdgeIds and getEdgeInfos when getting hidden edges to be consistant with AccumuloGraph

joeferner:


srfarley:
LGTM, assuming Travis succeeds.

PR #59, opened 2016-01-19
index element after altering visibility for accumulo

sfeng88:


core/src/main/java/org/vertexium/ElementBase.java
```diff
@@ -197,7 +197,7 @@ public long getTimestamp() {
         return timestamp;
     }
 
-    protected void setVisibility(Visibility visibility) {
+    public void setVisibility(Visibility visibility) {
```
joeferner: can we keep this protected and expose it only to Accumulo implementation as package private in AccumuloEdge and AccumuloVertex.

We really can't have people calling `setVisibility` on Elements.

PR #60, opened 2016-01-21
detect sql streaming properties length on write

srfarley:


sql/src/main/java/org/vertexium/sql/SqlStreamingPropertyTable.java
```diff
@@ -43,14 +46,20 @@ public StreamingPropertyValueRef put(String elementId, String key, String name,
...
                                 tableName, VALUE_COLUMN_NAME, VALUE_LENGTH_COLUMN_NAME, KEY_COLUMN_NAME),
-                        value.getInputStream(), value.getLength(), id);
+                        streamAndLength.inputStream, streamAndLength.length, id);
             }
             return new SqlStreamingPropertyValueRef(value, elementId, key, name, visibility, timestamp);
+        } finally {
+            try {
+                streamAndLength.inputStream.close();
+            } catch (IOException ex) {
+                Throwables.propagate(ex);
```
joeferner: can you throw a `VertexiumException` here instead with a message like "could not close input stream". I'm not a fan of `Throwables.propagate` since you end up getting a raw `RuntimeException`

sql/src/main/java/org/vertexium/sql/SqlStreamingPropertyTable.java
```diff
@@ -69,17 +78,30 @@ public StreamingPropertyValue get(String elementId, String key, String name, Vis
         }
     }
 
+    private static StreamAndLength streamAndLength(InputStream inputStream, long length) {
+        if (length < 0) {
+            try {
+                AutoDeleteFileInputStream fileInputStream = new AutoDeleteFileInputStream(inputStream);
+                return new StreamAndLength(fileInputStream, fileInputStream.getFileLength());
+            } catch (IOException ex) {
+                throw Throwables.propagate(ex);
```
joeferner: same here

srfarley: Done

sql/src/main/java/org/vertexium/sql/SqlStreamingPropertyTable.java
srfarley: Done.

sfeng88:
LGTM

PR #63, opened 2016-03-01
add method to get all historical values of all properties of an element

joeferner:


accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -745,14 +745,14 @@ public AccumuloNameSubstitutionStrategy getNameSubstitutionStrategy() {
                     String cq = column.getKey().getColumnQualifier().toString();
                     String columnVisibility = column.getKey().getColumnVisibility().toString();
                     if (column.getKey().getColumnFamily().equals(AccumuloElement.CF_PROPERTY)) {
-                        if (!columnVisibility.equals(visibility.getVisibilityString())) {
+                        if (visibility != null && !columnVisibility.equals(visibility.getVisibilityString())) {
```
sfeng88: By default is visibility "" or null?

joeferner: visibility is an object. The idea here is to skip this check if `null` is passed in. Similar to `name` and `key` below.

sfeng88:
LGTM

PR #65, opened 2016-03-02
SQL: create tables on startup if they do not exist. Add graph.createTables configuration to enable/disable this feature

joeferner:


sfeng88:
LGTM

PR #75, opened 2016-04-06
Efficiently delete an element property from its Elasticsearch document

srfarley:
This change submits an inline server-side Groovy script to remove the field from the document. Previously, the entire document was being re-added, which resulted in a merge, and thus did not remove the property from the index. This caused the element to be returned from queries that matched the deleted property value, and also affected total hit counts and aggregated results.

This is based on the Update API documentation: https://www.elastic.co/guide/en/elasticsearch/reference/1.7/docs-update.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v2.2.11
 
 * SQL: Change max primary key in DDL to 767 to fix MySQL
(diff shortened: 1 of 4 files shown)


sfeng88:
Can you add this fix to the changelog?

srfarley:
@sfeng88 Done

PR #79, opened 2016-04-22
Elasticsearch: only sort on properties know to be in the index

joeferner:
@sfeng88 @srfarley

### DIFF
--- a/test/src/main/java/org/vertexium/test/GraphTestBase.java
+++ b/test/src/main/java/org/vertexium/test/GraphTestBase.java
@@ -2085,6 +2085,23 @@ public abstract class GraphTestBase {
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


test/src/main/java/org/vertexium/test/GraphTestBase.java
```diff
@@ -2085,6 +2085,23 @@ public void testGraphQuerySort() {
     }
 
     @Test
+    public void testGraphQuerySortThatHasntBeenTested() {
```
srfarley: Did you mean to call this `testGraphQuerySortThatHasntBeenIndexed()`?

joeferner: fixed

sfeng88:
lgtm

srfarley:
👍

PR #82, opened 2016-04-27
add CalendarFieldAggregation

joeferner:
- [x] @sfeng88 
- [x] @srfarley 
- [x] @jharwig

### DIFF
--- a/core/src/main/java/org/vertexium/query/DefaultGraphQueryIterableWithAggregations.java
+++ b/core/src/main/java/org/vertexium/query/DefaultGraphQueryIterableWithAggregations.java
@@ -34,19 +34,26 @@ public class DefaultGraphQueryIterableWithAggregations<T extends Element> extend
         if (agg instanceof TermsAggregation) {
             return true;
         }
+        if (agg instanceof CalendarFieldAggregation) {
+            return true;
+        }
         return false;
     }
 
+    @SuppressWarnings("unchecked")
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/vertexium/query/DefaultGraphQueryIterableWithAggregations.java
```diff
@@ -55,7 +62,30 @@ public static boolean isAggregationSupported(Aggregation agg) {
...
+
+    private HistogramResult getCalendarFieldHistogramResult(final CalendarFieldAggregation agg, Iterator<T> it) {
+        String propertyName = agg.getPropertyName();
+        final Calendar calendar = GregorianCalendar.getInstance(agg.getTimeZone());
+        Map<Integer, List<T>> elementsByProperty = getElementsByProperty(it, propertyName, new ValueConverter<Integer>() {
+            @Override
+            public Integer convert(Object o) {
+                Date d = (Date) o;
+                calendar.setTime(d);
+                //noinspection MagicConstant
```
srfarley: Why is this needed? There are no constants referenced on this line.

joeferner: IntelliJ complains that there ISN'T a constant referenced on this line. It wants it to be a `Calendar` magic constant.

PR #86, opened 2016-05-03
Elasticsearch: fix: Aggregation after alter visibility

joeferner:
* [x] @sfeng88 
* [x] @srfarley


test/src/main/java/org/vertexium/test/GraphTestBase.java
```diff
@@ -4416,6 +4416,38 @@ public void testGraphQueryWithTermsAggregation() {
         assertEquals(1L, (long) edgePropertyCountByValue.get("label2"));
     }
 
+    @Test
+    public void testGraphQueryWithTermsAggregationAlterElementVisibility() {
```
sfeng88: Can we add a test for edges as well?

joeferner: done

PR #87, opened 2016-05-09
add table name into the context classpath variable to potentially hav…

rygim:
…e different iterators on the classpath

I know that there may be potential issues with classpaths potentially bleeding if there are updates to the iterators that are going to mismatch versions, but this is at least a start that will allow multiple user contexts to have their own iterators on the classpath and they won't overwrite each other's path's when different clients connect with different user names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v2.3.2
+ * ACCUMULO: iterator locations in accumulo config are now stored per table name
+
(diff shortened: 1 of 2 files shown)


joeferner:
@rygim Looks good. Can you add a line item to CHANGELOG.md.

rygim:
@joeferner oh yea, meant to do that.  Thanks for reminding me

PR #88, opened 2016-05-09
Elasticsearch: fix: sort by strings with tokens should not effect sort order

joeferner:
@sfeng88 @srfarley 

Sorting vertices with values "1-1" and "1-2" previously would not sort consistently because Elasticsearch would tokenize on `-`.

To avoid conflicts with existing "analyzed" sort fields I added a `_s` suffix to sorted fields (which I should have done in the first place anyway).

Can you think of some other unit tests I can add to verify the sort order is correct?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 # v2.3.2
- * ACCUMULO: iterator locations in accumulo config are now stored per table name
+
+* ACCUMULO: iterator locations in accumulo config are now stored per table name
+* Elasticsearch: fix: sort by strings with tokens should not effect sort order 
 
 # v2.3.1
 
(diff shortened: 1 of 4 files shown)


srfarley:
Other things to test:

* Add a property with VISIBILITY_C in `testGraphQuerySortOnPropertyWhichIsFullTextAndExactMatchIndexed` to insure it doesn't leak through.
* Test an exact match query ("1-2") and and a token match query ("1") to insure those queries still work in combination with sorting.

joeferner:
Added to test for exact match and token match.

Unfortunately because of the way we need to do sorting we are leaking sort order for properties you don't have access to. This PR doesn't change that. Unless we can come up with a different way to do sort I'm not sure if we can avoid it.

PR #92, opened 2016-05-17
optimize find paths for in-memory and sql implementations

srfarley:


srfarley:
@joeferner CHANGELOG is updated.

PR #95, opened 2016-05-23
Graph: add methods to get vertices and edges in a range of IDs

joeferner:
Accumulo: add methods to get the tablet splits

@srfarley

### DIFF
--- a/accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
+++ b/accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
@@ -4,7 +4,10 @@ import com.google.common.collect.Lists;
 import com.google.common.primitives.Longs;
 import org.apache.accumulo.core.client.*;
 import org.apache.accumulo.core.client.Scanner;
-import org.apache.accumulo.core.data.*;
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Mutation;
+import org.apache.accumulo.core.data.PartialKey;
+import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.iterators.IteratorUtil;
 import org.apache.accumulo.core.iterators.LongCombiner;
 import org.apache.accumulo.core.iterators.user.RowDeletingIterator;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/vertexium/GraphBase.java
```diff
@@ -114,6 +114,28 @@ protected boolean isIncluded(Vertex v) {
...
+    }
+
+    @Override
+    public Iterable<Vertex> getVerticesInRange(Range idRange, EnumSet<FetchHint> fetchHints, Authorizations authorizations) {
+        return getVerticesInRange(idRange, fetchHints, null, authorizations);
+    }
+
+    @Override
+    public Iterable<Vertex> getVerticesInRange(final Range idRange, EnumSet<FetchHint> fetchHints, Long endTime, Authorizations authorizations) {
+        LOGGER.warn("Performing scan of all vertices! Override getVerticesInRange.");
```
srfarley: I find these warning messages really annoying when running the unit tests. Can we just log each message once?

core/src/test/java/org/vertexium/RangeTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+        range = new Range(null, "c");
+        assertTrue(range.isInRange("b"));
+        assertFalse(range.isInRange("c"));
+
+        range = new Range("b", null);
+        assertFalse(range.isInRange("az"));
+        assertTrue(range.isInRange("b"));
+    }
+}
```
srfarley: There should be tests for `Range#equals()` and `Range#hashCode()`.

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -1546,6 +1591,42 @@ public Connector getConnector() {
         return connector;
     }
 
+    public Iterable<Range> listVerticesTableSplits() {
```
srfarley: Can these methods be package-protected?

joeferner: No, I need them in my spark code.

core/src/main/java/org/vertexium/GraphBase.java
joeferner: I agree. I'll make a new PR for that.

core/src/test/java/org/vertexium/RangeTest.java
joeferner: done

core/src/test/java/org/vertexium/RangeTest.java
```diff
@@ -0,0 +1,52 @@
...
+        assertFalse(range.isInRange("c"));
+
+        range = new Range("b", null);
+        assertFalse(range.isInRange("az"));
+        assertTrue(range.isInRange("b"));
+    }
+
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    @Test
+    public void testHashCode() {
```
srfarley: Yeah, maybe `hashCode` doesn't need tests.

joeferner: It does test the handling of nullness but that's about it.

PR #97, opened 2016-05-24
Remove excess logging

joeferner:
@srfarley 

I can't decide if I like this or just removing the warnings all together. Thoughts?

merge PR https://github.com/v5analytics/vertexium/pull/95 first


srfarley:
As an alternative, when the `Graph` subclass is created, we could use reflection to determine whether or not the appropriate methods were overridden, and log all of the warnings once from the constructor. Otherwise, this does seem like a lot of excess code, and it might be better to just remove the warnings.

joeferner:
I just removed all the messages. If you don't override you'll know it pretty quick when you get bad performance

PR #118, opened 2016-12-20
Added: support to read multiple StreamingPropertyValues with on query

joeferner:


core/src/main/java/org/vertexium/GraphBase.java
```diff
@@ -913,4 +914,13 @@ public void ensurePropertyDefined(String name, Object value) {
...
     }
+
+    @Override
+    public List<InputStream> getStreamingPropertyValueInputStreams(List<StreamingPropertyValue> streamingPropertyValues) {
+        List<InputStream> results = new ArrayList<>();
+        for (StreamingPropertyValue streamingPropertyValue : streamingPropertyValues) {
+            results.add(streamingPropertyValue.getInputStream());
+        }
+        return results;
+    }
```
mwizeman: This might be a little cleaner with streams.
```
        return streamingPropertyValues
                .stream()
                .map(StreamingPropertyValue::getInputStream)
                .collect(Collectors.toList());
```

mwizeman:
I just pushed a fix to the Kryo test that was causing Travis to fail. I'll check back in a bit to make sure Travis is passing.

PR #122, opened 2017-01-20
Fix: existing element mutations updating timestamps

joeferner:
Before this commit existing mutations would update the timestamp to now
losing the timestamp set in the mutations. This commit also fixes alter
edge label timestamps.

### DIFF
--- a/inmemory/src/main/java/org/vertexium/inmemory/InMemoryGraph.java
+++ b/inmemory/src/main/java/org/vertexium/inmemory/InMemoryGraph.java
@@ -308,6 +308,9 @@ public class InMemoryGraph extends GraphBaseWithSearchIndex {
     private Edge savePreparedEdge(final EdgeBuilderBase edgeBuilder, final String outVertexId, final String inVertexId, Long timestamp, Authorizations authorizations) {
         if (timestamp == null) {
             timestamp = IncreasingTime.currentTimeMillis();
+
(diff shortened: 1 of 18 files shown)


inmemory/src/main/java/org/vertexium/inmemory/InMemoryGraph.java
```diff
@@ -308,6 +308,11 @@ public Edge save(Authorizations authorizations) {
     private Edge savePreparedEdge(final EdgeBuilderBase edgeBuilder, final String outVertexId, final String inVertexId, Long timestamp, Authorizations authorizations) {
         if (timestamp == null) {
             timestamp = IncreasingTime.currentTimeMillis();
+
+            // The timestamps will be incremented below, this will ensure future mutations will be in the future
+            for (int i = 0; i < 10; i++) {
+                IncreasingTime.currentTimeMillis();
+            }
```
mwizeman: Maybe we should create a new method in `IncreasingTime` named something like `advanceLastTime`. The code itself would be a little more explicit about what's happening. This loop plus the bonus call in core/src/main/java/org/vertexium/mutation/AlterPropertyVisibility.java could be replaced with a call to the new method.

joeferner: done

PR #124, opened 2017-01-20
Added: support storing history in a separate table

joeferner:
By storing the history in a separate table we reduce the overhead of
skipping historic items for 99% of queries which only care about the
current values

### DIFF
--- a/accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
+++ b/accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
@@ -78,6 +78,7 @@ public class AccumuloGraph extends GraphBaseWithSearchIndex implements Traceable
     private final FileSystem fileSystem;
     private final String dataDir;
     private final CuratorFramework curatorFramework;
+    private final boolean historyInSeparateTable;
     private ThreadLocal<VertexiumMultiTableBatchWriter> elementWriter = new ThreadLocal<>();
     private ThreadLocal<BatchWriter> metadataWriter = new ThreadLocal<>();
     protected ElementMutationBuilder elementMutationBuilder;
@@ -86,7 +87,9 @@ public class AccumuloGraph extends GraphBaseWithSearchIndex implements Traceable
     private boolean foundVertexiumSerializerMetadata;
     private final AccumuloNameSubstitutionStrategy nameSubstitutionStrategy;
     private final String verticesTableName;
+    private final String historyVerticesTableName;
     private final String edgesTableName;
+    private final String historyEdgesTableName;
     private final String dataTableName;
     private final String metadataTableName;
     private final int numberOfQueryThreads;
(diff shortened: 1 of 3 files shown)


accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -1265,7 +1342,12 @@ ScannerBase createElementScanner(
             Authorizations authorizations
     ) throws VertexiumException {
         try {
-            String tableName = getTableNameFromElementType(elementType);
+            String tableName;
+            if (startTime != null || endTime != null || maxVersions == null || maxVersions > 1) {
```
mwizeman: Shouldn't this also check `isHistoryInSeparateTable()`

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -1628,19 +1737,35 @@ public Connector getConnector() {
         }
     }
 
+    public Iterable<Range> listHistoryVerticesTableSplits() {
+        try {
```
mwizeman: Consider making a private method `listTableSplits(String tableName)`? All of the various list*TableSplits methods would then be one-liner calls to that.

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -421,7 +443,21 @@ void softDeleteProperty(AccumuloElement element, Property property, Authorizatio
         }
     }
 
-    protected void addMutations(BatchWriter writer, Mutation... mutations) {
+    protected void addMutations(Element element, Mutation... mutations) {
+        _addMutations(getWriterFromElementType(element), mutations);
```
mwizeman: Body of this method could be replaced with:
```
addMutations(ElementType.getTypeFromElement(element), mutations);
```

joeferner: done

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -464,6 +508,38 @@ protected BatchWriter getWriterFromElementType(Element element) {
         }
     }
 
+    protected BatchWriter getHistoryWriterFromElementType(Element element) {
+        if (element instanceof Vertex) {
```
joeferner: done

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
joeferner: done

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
joeferner: done

PR #126, opened 2017-02-07
Added: size to TermsAggregation

joeferner:
Currently ElasticSearch defaults the number of results from a terms aggregation to 10. Should we up the default number of results or add this option or both?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v2.5.3
+
+* Added: size to TermsAggregation
+
 # v2.5.2
 
 * Added: support for Elasticsearch range aggregations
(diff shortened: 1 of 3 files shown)


sfeng88:
@joeferner I think we should do both. I can see a situation where certain customization would want the flexibility to change the default.

mwizeman:
I like the idea of making it configurable as well as changeable in code. 

In addition, while we're in updating this code I think we should also support the shard_size parameter as described [here in the docs](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/search-aggregations-bucket-terms-aggregation.html#_shard_size) and [here in the api](https://github.com/elastic/elasticsearch/blob/1.7/src/main/java/org/elasticsearch/search/aggregations/bucket/terms/TermsBuilder.java#L65). Adding support for this parameter allows the caller to have finer control over the accuracy without needing to increase the size returned to them.

joeferner:
How much of the elastic search insides do we want to expose? Should the users of vertexium realize they are using ES? What should the default size be?

PR #127, opened 2017-02-10
Added: Elasticsearch: Field query string support

joeferner:
This PR adds support for ES queries that have field names `name:joe`. With this change, future installations will require the installation of the Vertexium Elasticsearch plugin. As a nice side effect, instead of sending all the field names in the query, the query now looks like this:

```
"query" : {
  "vertexium_query_string" : {
    "query" : "Joe AND ferner",
    "authorizations" : [ "a" ]
  }
}
``` 

The plugin handles `vertexium_query_string` and expands `authorizations` into the fields that are supported.

cc @kunklejr


mwizeman:
Do you have notes anywhere on how to install the plugin into ES?

kunklejr:
Our docker build should obviously install this by default. Has that been done?

joeferner:
Tonight I'm going to add some code to make this optional so that we can get a Vertexium release out, then in a future version of Vertexium make it mandatory. This should allow older versions of software to still work without the plugin. I'll also add the instructions on adding the plugin.

mwizeman:
I have a new feature that I'm working on today that I would love to get into the next release. I hope to have PercentilesAggregation tested and finished up tonight.

mwizeman:
@kunklejr I can help update the docker image.

kunklejr:
@joeferner Do we need to wait until later to require the plugin? We're already going to force customers to upgrade their docker containers for release 3.1.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Changed: avoid saving empty mutations in Graph#saveElementMutations 
 * Added: support to read multiple StreamingPropertyValues with on query
 * Added: Graph.findPath exclusion labels
+* Added: Elasticsearch: Field query string support
```
mwizeman: This should be moved to the 2.5.3 section of the changelog as 2.5.2 has been released.

elasticsearch-singledocument-plugin/pom.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>vertexium-root</artifactId>
+        <groupId>org.vertexium</groupId>
+        <version>2.5.2-SNAPSHOT</version>
```
mwizeman: Update to 2.5.3-SNAPSHOT when #128 has been merged.

PR #128, opened 2017-02-24
Added: Elasticsearch: Support for Percentile Aggregations

mwizeman:
This adds support for ES percentile aggregations. Unlike other aggregations, we are unable to reduce the results of multiple percentile aggregations into a single answer. Because of this, a percentile aggregation must be run on a field for a specific visibility.

### DIFF
--- a/elasticsearch-singledocument/src/main/java/org/vertexium/elasticsearch/ElasticsearchSingleDocumentSearchIndex.java
+++ b/elasticsearch-singledocument/src/main/java/org/vertexium/elasticsearch/ElasticsearchSingleDocumentSearchIndex.java
@@ -593,7 +593,7 @@ public class ElasticsearchSingleDocumentSearchIndex implements SearchIndex, Sear
         return deflatePropertyName(graph, propertyName, propertyVisibility);
     }
 
-    private String deflatePropertyName(Graph graph, String propertyName, Visibility propertyVisibility) {
+    protected String deflatePropertyName(Graph graph, String propertyName, Visibility propertyVisibility) {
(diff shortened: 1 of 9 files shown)


elasticsearch-singledocument/src/main/java/org/vertexium/elasticsearch/ElasticsearchSingleDocumentSearchIndex.java
```diff
@@ -593,7 +593,7 @@ protected String deflatePropertyName(Graph graph, Property property) {
         return deflatePropertyName(graph, propertyName, propertyVisibility);
     }
 
-    private String deflatePropertyName(Graph graph, String propertyName, Visibility propertyVisibility) {
+    public String deflatePropertyName(Graph graph, String propertyName, Visibility propertyVisibility) {
```
joeferner: Can you change this to package scope instead of public

joeferner:
@mwizeman After those two changes you can go ahead and merge

mwizeman:
@joeferner Updated per your comments with 1a32e36230a5de6babfd9c705fb3bd52c865cbb6. When I was updating the changelog, I also noticed that the pom version for vertexium was still 2.5.2-SNAPHOT so I updated it to 2.5.3-SNAPSHOT.

joeferner:
Hmm, that's not good. I'll need to check on my Laptop to see how the version hasn't been updated.

sfeng88:
@joeferner Looking through the commit history, it doesn't have the maven release commits. Maybe it didn't get pushed?

PR #130, opened 2017-03-07
Added: ability to store extended data rows on an element

joeferner:
Extended data provides a way to store tabular data on an element yet
keep the initial element fetch lightweight. Each extended data row is
represented as a separate document in Elasticsearch to support advanced
searching.

I choose to not add metadata to extended data cells for performance reasons. Thoughts?


accumulo-iterators/src/main/java/org/vertexium/accumulo/iterator/util/DataOutputStreamUtils.java
```diff
@@ -26,6 +26,17 @@ public static void encodeTextList(DataOutputStream out, Collection<Text> texts)
         }
     }
 
+    public static void encodeStringSet(DataOutputStream out, Set<String> set) throws IOException {
```
diegogrz: Are empty sets possible? Should empty sets write out -1 as well?

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -57,6 +58,7 @@
 import static org.vertexium.util.IterableUtils.singleOrDefault;
 import static org.vertexium.util.IterableUtils.toList;
 import static org.vertexium.util.Preconditions.checkNotNull;
+import static org.vertexium.util.StreamUtils.stream;
 
 public class AccumuloGraph extends GraphBaseWithSearchIndex implements Traceable {
```
diegogrz: Should getHistoricalPropertyValues report one extended data?

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -2203,6 +2295,90 @@ protected GraphMetadataStore getGraphMetadataStore() {
...
+            ScannerBase scanner;
+            if (ranges == null || ranges.size() == 1) {
+                org.apache.accumulo.core.data.Range range = ranges == null ? null : ranges.iterator().next();
+                scanner = createScanner(tableName, range, authorizations);
+            } else {
+                scanner = createBatchScanner(tableName, ranges, authorizations);
+            }
+
+            IteratorSetting versioningIteratorSettings = new IteratorSetting(
+                    90,
```
diegogrz: Add variable name (or comment) to describe what the 90 (and 100 down below) values mean.

accumulo-iterators/src/main/java/org/vertexium/accumulo/iterator/util/DataOutputStreamUtils.java
joeferner: Empty sets will write out `0` and result in an empty set on the other side as opposed to `null` (-1)

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
joeferner: I'm not sure what to do with `getHistoricalPropertyValues`. I'm leaning toward not including extended data in that list until we have a need for it. Thoughts @mwizeman ?

core/src/main/java/org/vertexium/query/CompositeGraphQuery.java
```diff
@@ -106,6 +107,56 @@ protected boolean isIncluded(Element element) {
...
+    @Override
+    public QueryResultsIterable<? extends VertexiumObject> search(EnumSet<VertexiumObjectType> objectTypes, EnumSet<FetchHint> fetchHints) {
+        final Set<Object> seenIds = new HashSet<>();
+        return new QueryResultsSelectManyIterable<VertexiumObject>(this.queries) {
+            @Override
+            public Iterable<? extends VertexiumObject> getIterable(Query query) {
+                return query.search(objectTypes, fetchHints);
+            }
+
+            @Override
```
diegogrz: It is not very clear why a previously "seen" Id, should trigger the isIncluded flag to be false. Add comment or make SeenIds variable name a bit more descriptive.

accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
diegogrz: I agree. Mainly because it looks like it is not going to be that simple (since the data is in a different table). I just wanted to bring it up so we are all in agreement extended data is not part of regular history.

elasticsearch-singledocument/src/main/java/org/vertexium/elasticsearch/ElasticsearchSingleDocumentSearchIndex.java
```diff
@@ -718,16 +880,37 @@ private String getVisibilityHash(Graph graph, String propertyName, Visibility vi
...
-                        .request()
-        ).actionGet();
+        getClient().prepareDelete(indexName, ELEMENT_TYPE, id).execute().actionGet();
+    }
+
+    private void deleteExtendedDataForElement(Element element) {
+        try {
+            FilterBuilder filters[] = new FilterBuilder[1];
+            filters[0] = FilterBuilders.termFilter(ElasticsearchSingleDocumentSearchIndex.EXTENDED_DATA_ELEMENT_ID_FIELD_NAME, element.getId());
+            FilterBuilder filter = FilterBuilders.andFilter(filters);
```
diegogrz: Can we add a andFilter method that takes a single filter and calls the addFilter(filters[])?

elasticsearch-singledocument/src/main/java/org/vertexium/elasticsearch/score/ScoringStrategy.java
```diff
@@ -39,4 +40,12 @@ protected GraphConfiguration getGraphConfiguration() {
     public List<String> getFieldNames() {
         return new ArrayList<>();
     }
+
```
diegogrz: No change needed here but, Is there any particular reason why only certain methods are required to be defined by ScoringStrategy(s) implementations?  Just a bit concern that someone may develop a ScoringStrategy not implementing those methods, and a user of that strategy downstream uses them (because they are available) expecting them to do something.

joeferner:
don't merge quite yet

PR #134, opened 2017-03-23
Changed: Remove property metadata from default fetch hints

joeferner:
Bringing property metadata back is expensive as the metadata grows and
most times you don't need this information. This changes the default to
not bring metadata back.

### DIFF
--- a/accumulo/src/main/java/org/vertexium/accumulo/mapreduce/AccumuloEdgeInputFormat.java
+++ b/accumulo/src/main/java/org/vertexium/accumulo/mapreduce/AccumuloEdgeInputFormat.java
@@ -15,7 +15,6 @@ import org.vertexium.accumulo.AccumuloEdge;
 import org.vertexium.accumulo.AccumuloGraph;
 import org.vertexium.accumulo.iterator.EdgeIterator;
 import org.vertexium.accumulo.iterator.model.EdgeElementData;
-import org.vertexium.accumulo.iterator.model.FetchHint;
 import org.vertexium.mutation.PropertyDeleteMutation;
 import org.vertexium.mutation.PropertySoftDeleteMutation;
 
@@ -24,7 +23,7 @@ import java.util.EnumSet;
 import java.util.Map;
 
 public class AccumuloEdgeInputFormat extends AccumuloElementInputFormatBase<Edge> {
-    private static EdgeIterator edgeIterator = new EdgeIterator(AccumuloGraph.toIteratorFetchHints(org.vertexium.FetchHint.ALL));
+    private static EdgeIterator edgeIterator = new EdgeIterator(AccumuloGraph.toIteratorFetchHints(FetchHint.DEFAULT));
 
     public static void setInputInfo(Job job, AccumuloGraph graph, String instanceName, String zooKeepers, String principal, AuthenticationToken token, String[] authorizations) throws AccumuloSecurityException {
         String tableName = graph.getEdgesTableName();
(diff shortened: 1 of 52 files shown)


accumulo/src/main/java/org/vertexium/accumulo/mapreduce/AccumuloEdgeInputFormat.java
```diff
@@ -24,23 +23,27 @@
...
     @Override
-    protected Edge createElementFromRow(AccumuloGraph graph, PeekingIterator<Map.Entry<Key, Value>> row, Authorizations authorizations) {
+    protected Edge createElementFromRow(
+            AccumuloGraph graph,
+            PeekingIterator<Map.Entry<Key, Value>> row,
+            Authorizations authorizations
+    ) {
         try {
-            EnumSet<FetchHint> fetchHints = AccumuloGraph.toIteratorFetchHints(org.vertexium.FetchHint.ALL);
+            EnumSet<org.vertexium.FetchHint> fetchHints = org.vertexium.FetchHint.DEFAULT;
```
mwizeman: Why use FQN here rather than importing it?

accumulo/src/main/java/org/vertexium/accumulo/mapreduce/AccumuloVertexInputFormat.java
```diff
@@ -39,13 +38,13 @@ protected Vertex createElementFromRow(AccumuloGraph graph, PeekingIterator<Map.E
 
     public static Vertex createVertex(AccumuloGraph graph, Iterator<Map.Entry<Key, Value>> row, Authorizations authorizations) {
         try {
-            EnumSet<FetchHint> fetchHints = AccumuloGraph.toIteratorFetchHints(org.vertexium.FetchHint.ALL);
+            EnumSet<org.vertexium.FetchHint> fetchHints = org.vertexium.FetchHint.DEFAULT;
```
mwizeman: Same FQN comment.

core/src/main/java/org/vertexium/FetchHint.java
```diff
@@ -60,4 +61,18 @@ public static String toString(EnumSet<FetchHint> fetchHints) {
             return EnumSet.copyOf(results);
         }
     }
+
+    public static void checkFetchHints(EnumSet<FetchHint> fetchHints, FetchHint neededFetchHint) {
+        if (!fetchHints.contains(neededFetchHint)) {
```
mwizeman: Small change. This method could be implemented as:
```
checkFetchHints(fetchHints, EnumSet.of(neededFetchHint))
```
I'd be fine if they stay the way they are though.

accumulo/src/main/java/org/vertexium/accumulo/mapreduce/AccumuloEdgeInputFormat.java
joeferner: There are two `FetchHint` classes. One for iterators and one in Vertexium core. I just renamed the iterator fetch hint to `IteratorFetchHint` and fixed the FQNs

accumulo/src/main/java/org/vertexium/accumulo/mapreduce/AccumuloVertexInputFormat.java
joeferner: same here

core/src/main/java/org/vertexium/FetchHint.java
joeferner: Changed

PR #137, opened 2017-03-28
When find paths max 2 hops away, return paths one hop away as well

sfeng88:
Don't merge this PR. Want to see if test passes in Travis.


joeferner:
@sfeng88  This looks, as soon as travis passes go ahead and merge

PR #142, opened 2017-03-31
Accumulo + Elasticsearch query with FetchHint.NONE

### LINKED ISSUE #135: When querying for a vertex with FETCH_HINT.NONE, accumulo graph doesn't return the vertex, but sql graph does

mwizeman:
This pull requests includes a few related changes:

1. Refactored the Accumulo, Elasticsearch, and Elasticsearch2 tests to make their embedded servers a little easier to use from other modules.
2. Added a new module named `vertexium-multimodule-test` that tests the combinations of Accumulo + Elasticsearch and Accumulo + Elasticsearch2
3. Added a test to GraphTestBase that verified issue #135 in both of the new test modules
4. Fixed #135 with updates to `ElasticSearchSingleDocumentSearchQueryBase.java` and `ElasticsearchSearchQueryBase.java`
5. Deprecated `EdgeCountScoringStrategy`


accumulo/src/test/java/org/vertexium/accumulo/AccumuloResource.java
```diff
@@ -0,0 +1,148 @@
...
+import org.vertexium.util.VertexiumLoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.vertexium.test.GraphTestBase.*;
+
+public class AccumuloResource extends ExternalResource {
```
joeferner: Not really a ding on this PR but more of a discussion point: With this new `AccumuloResource` strategy could we break up the way to long `GraphTestBase` class?

CC @diegogrz

elasticsearch-singledocument/src/test/resources/log4j.xml
```diff
@@ -14,10 +14,6 @@
         </layout>
     </appender>
 
-    <logger name="org.vertexium.query.Query.QUERY" additivity="false">
```
joeferner: Instead of removing these lines should we comment them out with a comment stating something about to debug query you can turn this on or something. Or probably better yet put it in the ES readme. I always end up having to look up how to do this.

elasticsearch2/src/test/resources/log4j.xml
```diff
@@ -26,10 +26,6 @@
         <level value="DEBUG" />
         <appender-ref ref="console" />
     </logger>
-    <logger name="org.vertexium.query.Query.QUERY" additivity="false">
```
joeferner: same as above, comment out or readme.

multimodule-test/accumulo-elasticsearch/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.vertexium</groupId>
+            <artifactId>vertexium-accumulo</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
joeferner: By including these `test-jar`s does it run the Accumulo and ES tests from those again?

multimodule-test/accumulo-elasticsearch2/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,33 @@
...
+>
+
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
+    <appender name="console" class="org.vertexium.elasticsearch2.utils.ConsoleAppender">
+        <layout class="org.apache.log4j.PatternLayout">
+            <param name="ConversionPattern" value="&pattern;" />
+        </layout>
+    </appender>
+
+    <logger name="org.vertexium.query.Query.QUERY" additivity="false">
```
joeferner: Should this be removed to be consistent?

elasticsearch-singledocument/src/test/resources/log4j.xml
mwizeman: This logger was actually the source of my log4j issue yesterday. If you look 3 loggers down in this file, you'll find that it has a duplicate. I just removed the extra one to make log4j happy.

elasticsearch2/src/test/resources/log4j.xml
mwizeman: Same as above, though I apparently chose to remove the second one instead of the first in this file :-)

multimodule-test/accumulo-elasticsearch/pom.xml
mwizeman: I'm pretty sure it doesn't, but I will double check that. This dependency is so that I could get at the AccumuloResource and ElasticsearchResource. I'll review the Travis log and comment again to verify that it doesn't rerun them.

multimodule-test/accumulo-elasticsearch2/src/test/resources/log4j.xml
mwizeman: Nope. There is only one QUERY logger in this file.

mwizeman: Nope. There is only one QUERY logger in this file.

elasticsearch-singledocument/src/test/resources/log4j.xml
mwizeman: This logger was actually the source of my log4j issue yesterday. If you look 3 loggers down in this file, you'll find that it has a duplicate. I just removed the extra one to make log4j happy.

elasticsearch2/src/test/resources/log4j.xml
mwizeman: Same as above, though I apparently chose to remove the second one instead of the first in this file :-)

multimodule-test/accumulo-elasticsearch/pom.xml
mwizeman: I'm pretty sure it doesn't, but I will double check that. This dependency is so that I could get at the AccumuloResource and ElasticsearchResource. I'll review the Travis log and comment again to verify that it doesn't rerun them.

mwizeman: I just pulled up the Travis log and searched for `AccumuloSimpleSubstitutionGraphTest` to verify that it was only run a single time. I think we're good.

accumulo/src/test/java/org/vertexium/accumulo/AccumuloResource.java
diegogrz: @joeferner @mwizeman I did split the GraphTestBase into multiple tests but I did not push it through yet because it was kind of ugly and I was hoping to find a better way. I'll try to summarize. I split GraphTestBase into multiple "suites" that could be called from "GraphTestBase ". That works fine but the "suites" are meant to be independent. In order to share a common environment, you have to make the suites refer to a "static" environment that has the graph, authorizations, etc. However, those members are defined at instantiation time (i.e. different for Accumulo, InMemory, etc). Since you cant override static functions in Java (for a valid philosophical reason), I kinda had to make an Common Environment class that gets instantiated by the GraphTest subclasses but overrides a static variable than the suites can access to. It works but I dont feel too happy about overriding static member once a class has been instantiated (seems wrong). I can push my changes to my branch (they are on my machine at home) and let you guys look at it to see what I am talking about. I couldnt think of a better way to split GraphTests without major refactoring.

mwizeman: Using the ClassRules seems to be a good way to pull out common BeforeClass/AfterClass and Before/After behavior from tests. I haven't played with them at all, but could we use [JUnit 5 dynamic tests](http://junit.org/junit5/docs/current/user-guide/#writing-tests-dynamic-tests) to somehow break the graph tests in to parts that can be included at test time?

mwizeman:
@joeferner Squashed. I'll let Travis finish up the build and then merge. Thanks!

accumulo/src/test/java/org/vertexium/accumulo/AccumuloResource.java
diegogrz: @mwizeman I did play around with them and they work fine but I believe the class has to be static. Not sure we can make it work with abstract GraphTestBase.

diegogrz: @mwizeman I'll read on JUnit5 dynamic tests tonight.

PR #143, opened 2017-04-03
ability to return if there is any paths between two vertices

sfeng88:


core/src/main/java/org/vertexium/FindPathOptions.java
```diff
@@ -17,6 +18,20 @@ public FindPathOptions(String sourceVertexId, String destVertexId, int maxHops)
...
         this.destVertexId = destVertexId;
         this.maxHops = maxHops;
+        this.getAnyPath = false;
+    }
+
+    /**
+     * @param sourceVertexId The source vertex id to start the search from.
+     * @param destVertexId   The destination vertex id to get to.
+     * @param maxHops        The maximum number of hops to make before giving up.
+     * @param getAnyPath     Return true if a path is found
```
mwizeman: Maybe the description of this param should be something more like `return as soon as the first path is found`

sfeng88: done.

joeferner:
@sfeng88 I'm seeing a timeout on Travis, don't think it has to do with your changes though

mwizeman:
@joeferner It looks like travis-ci.org has a 50 minute timeout. With the addition of the multi-module tests, we're right at the limit. I just pushed an update to `.travis.yml` and `pom.xml` on this branch to split the build into two jobs, one for all of the individual module tests and one for the multi-module tests. The jobs run in parallel so the build now [completes in 26 minutes](https://travis-ci.org/visallo/vertexium/builds/218217409). According to the [travis docs](https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts) the 50 minutes is per-job so hopefully this helps.

PR #145, opened 2017-04-24
Changed: Removed timestamp from streaming property value row key.

joeferner:
With the timestamp in the SPV row key it prevents the cleanup of history.

### DIFF
--- a/accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
+++ b/accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
@@ -1,6 +1,8 @@
 package org.vertexium.accumulo;
 
-import com.google.common.collect.*;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.primitives.Longs;
 import org.apache.accumulo.core.client.*;
 import org.apache.accumulo.core.client.Scanner;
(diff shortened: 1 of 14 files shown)


accumulo/src/main/java/org/vertexium/accumulo/AccumuloGraph.java
```diff
@@ -1651,6 +1654,16 @@ public Edge getEdge(String edgeId, EnumSet<FetchHint> fetchHints, Long endTime,
 
             final long timerStartTime = System.currentTimeMillis();
             ScannerBase scanner = createBatchScanner(getDataTableName(), ranges, new org.apache.accumulo.core.security.Authorizations());
+            if (timestamp != null) {
```
joeferner: This change is a bit scary and I'm not sure how to verify unless it's tested in a real world scenario. I can image that it could be possible that a SPV was written in the past and now that the timestamp is not in the key I need to look it up based on the property's timestamp, which could be off? I'm not really sure.

joeferner:
Don't merge this yet I'm going to test this on real data

accumulo/src/main/java/org/vertexium/accumulo/keys/DataTableRowKey.java
```diff
@@ -41,7 +44,10 @@ public String getPropertyKey() {
         return parts[PARTS_INDEX_PROPERTY_KEY];
     }
 
-    public long getPropertyTimestamp() {
-        return Long.parseLong(parts[PARTS_INDEX_PROPERTY_TIMESTAMP]);
+    public static boolean isLegacy(String dataRowKey) {
```
joeferner: @mwizeman based on problems I had on real data I need to check the following and skip the timestamp limiting above.

PR #146, opened 2017-04-26
Changed: The default behavior of calling methods without fetch hints …

joeferner:
…will use the default fetch hints specified in the graph configuration

To move 2.6 forward I change the default fetch hints to be configurable. The default will use the old behavior of returning the property metadata.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Added: ability to return if there is any paths between two vertices
 * Added: ability to store extended data rows on an element
 * Added: Elasticsearch 2.x support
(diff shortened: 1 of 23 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Added: ability to return if there is any paths between two vertices
 * Added: ability to store extended data rows on an element
 * Added: Elasticsearch 2.x support
+* Changed: The default behavior of calling methods without fetch hints will use the default fetch hints specified in the graph configuration
```
mwizeman: Maybe mention that since the default is now configurable, FetchHint.DEFAULT was removed. I can't imagine it would appear very frequently in people's code, but if it does it might be good to use this `Changed` description to let them know why.

PR #147, opened 2017-04-26
Added: Cypher query support

joeferner:
Based on the http://www.opencypher.org/ spec


cli/src/main/java/org/vertexium/cli/VertexiumShell.java
```diff
@@ -121,7 +128,16 @@ public Object call(Object... args) {
...
-        startGroovysh(evalString, fileNames);
+        Groovysh shell = createShell();
+        shell.execute("import " + Visibility.class.getPackage().getName() + ".*;");
+        shell.execute("v = new " + LazyVertexMap.class.getName() + "();");
+        shell.execute("e = new " + LazyEdgeMap.class.getName() + "();");
+        shell.execute("cypher = new " + LazyEdgeMap.class.getName() + "();");
+        shell.execute("g = " + VertexiumScript.class.getName() + ".getGraph();");
+        shell.execute("auths = " + VertexiumScript.class.getName() + ".getAuthorizations();");
+        shell.execute("time = " + VertexiumScript.class.getName() + ".getTime();");
+        shell.execute("cypher = { code -> " + VertexiumScript.class.getName() + ".executeCypher(code) };");
```
mwizeman: Doesn't this variable name collide with the one on line 135?

core/src/main/java/org/vertexium/ElementBase.java
```diff
@@ -525,4 +525,12 @@ protected boolean isIncluded(HistoricalPropertyValue pv) {
     public EnumSet<FetchHint> getFetchHints() {
         return fetchHints;
     }
+
+    @Override
+    public int compareTo(Object o) {
```
mwizeman: Since you've provided a default implementation in the interface, is this needed?

cli/src/main/java/org/vertexium/cli/VertexiumShell.java
joeferner: oops. done

core/src/main/java/org/vertexium/ElementBase.java
joeferner: done

PR #149, opened 2017-05-02
Added: tool to delete historical legacy SPV data

joeferner:
This is a pretty critical script. It will delete the old SPV data that has timestamps in the key if there exists a SPV without the timstamp.

### DIFF
--- a/accumulo/src/main/java/org/vertexium/accumulo/tools/DeleteHistoricalLegacyStreamingPropertyValueData.java
+++ b/accumulo/src/main/java/org/vertexium/accumulo/tools/DeleteHistoricalLegacyStreamingPropertyValueData.java
@@ -1,0 +1,139 @@
+package org.vertexium.accumulo.tools;
+
+import org.apache.accumulo.core.client.BatchWriter;
+import org.apache.accumulo.core.client.MutationsRejectedException;
+import org.apache.accumulo.core.client.Scanner;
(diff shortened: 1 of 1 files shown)


accumulo/src/main/java/org/vertexium/accumulo/tools/DeleteHistoricalLegacyStreamingPropertyValueData.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+        } catch (Exception ex) {
+            throw new VertexiumException("Could not delete old SPV data", ex);
+        }
+    }
+
+    private void deleteRows(BatchWriter writer, List<Key> rowsToDelete, Options options) throws MutationsRejectedException {
+        rowsToDelete.sort(Comparator.comparingLong(Key::getTimestamp));
+        int i = 0;
+        for (Key key : rowsToDelete) {
```
mwizeman: Not really a problem, but would this loop and counting get any cleaner if you used a while loop and popped the values off the list? I'm thinking something like:
```
while (rowsToDelete.size() > options.getVersionsToKeep()) {
  Key key = rowsToDelete.remove(0);
  LOGGER.debug("deleting row: %s", key.getRow().toString());
  if (!options.isDryRun()) {
  ...
}
```
If you don't want to modify the contents of your parameter, you could put it in a new List first.

PR #157, opened 2017-08-08
remove deprecated interfaces, examples, and change es2 deprecated methods

joeferner:
This is the first step to getting ES5 working. The ES deprecated methods were removed in ES5. Removing the deprecated interfaces and examples should also makes the transition easier.

The unit tests passed on my machine but we'll need to keep an eye on CI.


elasticsearch2-plugin/src/main/java/org/vertexium/elasticsearch2/plugin/VertexiumMapperQueryParser.java
```diff
@@ -37,16 +39,17 @@ protected Query newFieldQuery(Analyzer analyzer, String field, String queryText,
         }
 
         String fieldPrefix = field + "_";
-        DisjunctionMaxQuery query = new DisjunctionMaxQuery(0.0f);
+        List<Query> disjucts = new ArrayList<>();
```
mwizeman: Typo?  maybe `disjuncts`

joeferner: fixed

PR #161, opened 2017-09-07
If only a string is needed, use decodeString rather than decodeText

mwizeman:
Using decode Text when only a string is needed causes unnecessary garbage `Text` and `char[]` objects.


sfeng88:
@mwizeman should we add an entry to the changelog for this optimization?

kunklejr:
I wouldn't bother adding anything to the changelog unless the performance improvement is really dramatic.

mwizeman:
@sfeng88 I didn't add anything to the changelog since it didn't change any behavior that downstream projects would notice.

@kunklejr In my quick testing, it saved me about 7-8% on object garbage collection. Not a dramatic increase that I think would warrant mentioning.

PR #163, opened 2017-09-11
implement does not contain for text

sfeng88:


test/src/main/java/org/vertexium/test/GraphTestBase.java
```diff
@@ -3752,6 +3752,33 @@ public void testTextIndex() throws Exception {
...
+
+        graph.prepareVertex("v3", VISIBILITY_A)
+                .setProperty("both", "Temp", VISIBILITY_A)
+                .save(AUTHORIZATIONS_A_AND_B);
+
+        graph.prepareVertex("v4", VISIBILITY_A)
+                .save(AUTHORIZATIONS_A_AND_B);
+        graph.flush();
+
+        Assert.assertEquals(3, count(graph.query(AUTHORIZATIONS_A).has("both", TextPredicate.DOES_NOT_CONTAIN, "Test").vertices()));
```
mwizeman: @sfeng88 For each assert statement here, could you add another line to check which three matched?

sfeng88:
@mwizeman I modified the test based on your comment. 

I also found a bug in my original code where the elasticsearch total hit count differed from the number of vertices being returned. (Fixed: https://github.com/visallo/vertexium/pull/163/commits/b63169a0e13cde57ab6e098b7260606a67b1ea65)

sfeng88:
@mwizeman @joeferner are you guys ok with this change?

PR #164, opened 2017-09-12
added es shard size configuration for term aggregations

sfeng88:


joeferner:
Travis errors but I'm assuming they are incorrect can you take a look

sfeng88:
@joeferner it looks like a timeout issue with travis. I'll keep this as a PR and run the test locally one more time to make sure. If it passes locally, I'll merge the PR.

PR #166, opened 2017-09-21
Support Elasticsearch 5

mwizeman:
Drop support for Elasticsearch 2 and add support for Elasticsearch 5.


elasticsearch5/pom.xml
```diff
@@ -20,49 +20,96 @@
...
-            <groupId>org.codehaus.groovy</groupId>
-            <artifactId>groovy-all</artifactId>
-            <version>2.4.4</version>
+            <groupId>org.vertexium</groupId>
+            <artifactId>vertexium-elasticsearch5-plugin</artifactId>
+            <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <!--<dependency>-->
```
joeferner: remove

joeferner:
can we also rebase and merge some of the commits

elasticsearch5/pom.xml
mwizeman: Removed

mwizeman:
@joeferner rebased and squashed all of the commits

joeferner:
@mwizeman I'm going to make one last release of 2.6 tonight. After that are we ready to merge and cut a 3.x release?

mwizeman:
@joeferner I think so. I pushed the strict query checking and multi-fields today.

sfeng88:
@joeferner you're releasing [2.4.6](https://github.com/visallo/vertexium/tree/2.4.x) tonight right?

joeferner:
@sfeng88 both 2.4.x and 2.6.x

core/src/main/java/org/vertexium/query/QueryBase.java
```diff
@@ -351,6 +351,11 @@ public HasValueContainer(final String key, final Predicate predicate, final Obje
             this.value = value;
             this.predicate = predicate;
             this.propertyDefinitions = propertyDefinitions;
+
+            PropertyDefinition propertyDefinition = PropertyDefinition.findPropertyDefinition(propertyDefinitions, key);
+            if (!predicate.isSupported(propertyDefinition)) {
```
joeferner: Instead of `isSupported` should we make this something like `validate(propertyDefinition)` that way we can through a more meaningful error message, for example in the case below we can say you are missing full text hint.

mwizeman: @joeferner Good idea! Changed with 1cf5098361e7756397ca838422294ed2866ccc63.

PR #167, opened 2017-09-28
fix does not contain return wrong count for es

sfeng88:


joeferner:
@sfeng88 Looks good, need to fix the merge conflicts

PR #170, opened 2017-09-29
DOES_NOT_CONTAIN queries should use AND between property names

mwizeman:
It looks like this was already fixed in the ElasticSearch 1.7 module but missed in the new ElasticSearch 5 code.

### DIFF
--- a/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
+++ b/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
@@ -756,6 +756,9 @@ public class ElasticsearchSearchQueryBase extends QueryBase {
                     throw new VertexiumException("Unexpected text predicate " + has.predicate);
             }
         }
+        if (compare.equals(TextPredicate.DOES_NOT_CONTAIN)) {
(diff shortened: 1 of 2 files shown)


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
```diff
@@ -756,6 +756,9 @@ protected QueryBuilder getFilterForTextPredicate(TextPredicate compare, HasValue
                     throw new VertexiumException("Unexpected text predicate " + has.predicate);
             }
         }
+        if (compare.equals(TextPredicate.DOES_NOT_CONTAIN)) {
```
joeferner: I swear this is the change @sfeng88 made earlier

mwizeman: It looks like it was only made in the ES 1.7 module in the previous PR.

sfeng88: The tests definitely passed for me last night...does the ES5 test have a new profile? that might be it.

PR #175, opened 2017-10-27
Add a hasId method to the Query class

mwizeman:
This PR adds the ability to filter a query by id, which improves the performance of multivertex queries.

When [running a MultiVertexQuery](https://github.com/visallo/vertexium/blob/master/core/src/main/java/org/vertexium/Graph.java#L760-L777), the [ElasticSearch code was using the DefaultMultiVertexQuery](https://github.com/visallo/vertexium/blob/master/elasticsearch-singledocument/src/main/java/org/vertexium/elasticsearch/ElasticsearchSingleDocumentSearchIndex.java#L1387-L1389) which applies all of the search filters after loading all of the elements. This PR will improve performance by filtering the elements in Elasticsearch before loading them from the datastore.

### DIFF
--- a/core/src/main/java/org/vertexium/query/Query.java
+++ b/core/src/main/java/org/vertexium/query/Query.java
@@ -76,7 +76,7 @@ public interface Query {
      * @param ids The ids to filter on.
      * @return The query object, allowing you to chain methods.
      */
-    Query hasId(Collection<String> ids);
+    Query hasId(Iterable<String> ids);
 
     /**
      * Adds a edge label filter to the query.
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/vertexium/query/Query.java
```diff
@@ -63,6 +63,22 @@
...
+     */
+    Query hasId(String... ids);
+
+    /**
+     * Adds an id filter to the query.
+     *
+     * @param ids The ids to filter on.
+     * @return The query object, allowing you to chain methods.
+     */
+    Query hasId(Collection<String> ids);
```
joeferner: Can you make this `Iterable` instead of `Collection`

mwizeman: Fixed with https://github.com/visallo/vertexium/pull/175/commits/095e4883c02e595c40723a876504f7da2fd80ced

PR #180, opened 2017-11-01
Queue Elasticsearch field removal

mwizeman:
In the Elasticsearch plugins, adding/updating documents to the index is queued as a future but field removal was immediate. This update also queues field removal to avoid timing issues.

In particular, there was a problem in [AccumuloElement#saveExistingElementMutation](https://github.com/visallo/vertexium/blob/14f748037bf3ec0081324f7cff7a483aecc0cfa6/accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java#L128) if there were property changes as well as an element visibility update.  The ES document was [updated on this line](https://github.com/visallo/vertexium/blob/14f748037bf3ec0081324f7cff7a483aecc0cfa6/accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java#L146). The elementType property was removed and then the ES document was updated again [on this line](https://github.com/visallo/vertexium/blob/14f748037bf3ec0081324f7cff7a483aecc0cfa6/accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java#L156). Logically, the order should have been: "update doc", "remove field", "update doc". Because updates were queued and removals weren't, the actual order was _somtimes:_ "remove field", "update doc", "update doc".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Changed: Reduced DefaultIndexSelectionStrategy cache load time from 1hr to 5min
 * Added: Added a hasId method to the Query class to allow searches to be filtered by element ID.
 * Fix: Extended data element type value for edges
+* Changed: Field removal from Elasticsearch documents is now queued as a future instead of immediate
 
 # v3.0.0
 * Changed: Removed ES 2 support and replaced it with ES 5 support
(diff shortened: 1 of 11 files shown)


joeferner:
Stupid Travis

mwizeman:
@joeferner I updated the tests to make sure that each Elasticsearch has a distinct cluster name and that seemed to help.

joeferner:
awsome, looks like it passed, now just need to fix the conflict, and you can merge

PR #182, opened 2017-11-06
update es5 configuration for in process node

sfeng88:


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchIndexConfiguration.java
```diff
@@ -42,6 +42,7 @@
     public static final String IN_PROCESS_NODE_DATA_PATH = "inProcessNode.dataPath";
```
joeferner: Are these other paths needed anymore?

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -135,22 +135,13 @@ protected Client createClient(ElasticsearchSearchIndexConfiguration config) {
...
-            checkNotNull(logsPath, ElasticsearchSearchIndexConfiguration.IN_PROCESS_NODE_LOGS_PATH + " is required for in process Elasticsearch node");
-            String workPath = config.getInProcessNodeWorkPath();
-            checkNotNull(workPath, ElasticsearchSearchIndexConfiguration.IN_PROCESS_NODE_WORK_PATH + " is required for in process Elasticsearch node");
-            int numberOfShards = config.getNumberOfShards();
+            String homePath = config.getInProcessNodeHomePath();
+            checkNotNull(homePath, ElasticsearchSearchIndexConfiguration.IN_PROCESS_NODE_HOME_PATH + " is required for in process Elasticsearch node");
 
             Map<String, String> mapSettings = new HashMap<>();
-            mapSettings.put("script.inline", "true");
-            mapSettings.put("index.number_of_shards", Integer.toString(numberOfShards));
```
joeferner: How do I configure this? I kinda need this.

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -135,22 +135,13 @@ protected Client createClient(ElasticsearchSearchIndexConfiguration config) {
...
-            mapSettings.put("index.number_of_replicas", "0");
-            mapSettings.put("path.data", dataPath);
-            mapSettings.put("path.logs", logsPath);
-            mapSettings.put("path.work", workPath);
-            mapSettings.put("discovery.zen.ping.multicast.enabled", "false");
```
joeferner: Is Zen disabled by default?

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchIndexConfiguration.java
sfeng88: The other paths are needed for ES 1.7. In ES 5, the three paths are combined into `path.home`.

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
sfeng88: In ES5, index level settings can't be on node level configurations. The properties in the configuration file are still the same. We set the values on the creation of indices: https://github.com/visallo/vertexium/blob/master/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java#L1853-L1856

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
sfeng88: unicast is the default now and `discovery.zen.ping.multicast.enabled` is no longer a valid config in ES5

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchIndexConfiguration.java
sfeng88: @mwizeman just pointed out this is in the ES5 package so those configs are not needed. just pushed the deletion

PR #183, opened 2017-11-06
 Fix Timing Issue and Potential Infinite Loop

mwizeman:
`PagingIterable`: Fixed an infinite looping problem in the PagingIterable that may have resulted from a malformed ES document.  

`AccumuloElement`: Fixed a timing issue in `AccumuloElement.saveExistingElementMutation` that resulted in an ES document having both the old and new visibility. Since the update `Future`s happen in parallel, it's important to ensure that the removal of the old permission happens after the updating of the properties.

`Elasticsearch5SearchIndex`: Made a small performance improvement in `Elasticsearch5SearchIndex.removeFieldsFromDocument` that will not bother running the script in Elasticsearch if there are no properties to remove.

`ElasticsearchSingleDocumentSearchIndex`: Same improve as described above for `Elasticsearch5SearchIndex`

### DIFF
--- a/accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java
+++ b/accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java
@@ -153,6 +153,7 @@ public abstract class AccumuloElement extends ElementBase implements Serializabl
         );
 
         if (mutation.getNewElementVisibility() != null) {
+            getGraph().flush();
             getGraph().alterElementVisibility((AccumuloElement) mutation.getElement(), mutation.getNewElementVisibility(), authorizations);
         }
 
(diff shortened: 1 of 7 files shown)


accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java
```diff
@@ -153,6 +153,7 @@ protected void setVisibility(Visibility visibility) {
         );
 
         if (mutation.getNewElementVisibility() != null) {
+            getGraph().flush();
```
sfeng88: can we add a comment for why we added this graph.flush here to remind us of the reason?

joeferner: Can you add a comment on this line describing why it's needed

mwizeman: @sfeng88 Done with https://github.com/visallo/vertexium/pull/183/commits/3934f54bb4dd96a8300bb5a171a9d37e2cb9d9a4

mwizeman: @joeferner Done with https://github.com/visallo/vertexium/pull/183/commits/3934f54bb4dd96a8300bb5a171a9d37e2cb9d9a4

PR #184, opened 2017-11-07
GeoShape Indexing Performance Improvement

mwizeman:
Indexing large GeoShape properties (large in terms of land area) was slow. I found a few Elasticsearch issues (https://github.com/elastic/elasticsearch/issues/25844, https://github.com/elastic/elasticsearch/issues/25833#issuecomment-317116918, https://github.com/elastic/elasticsearch/issues/22087#issuecomment-334759601) discussing the problem and this PR incorporates the advice from there. This improved the index time for my use case from ~20s to ~500ms.


joeferner:
@mwizeman rebase please, changelog

mwizeman:
@joeferner Done with the rebase.

andrecordeiro:
@mwizeman Hello, I have a question. I mounted a regionmap, inside I have polygons and multipológonos, which are created based on my .geojson file that is configured in my kibana.yml file. I do join with my index in elastic, to insert only one information in the polygon. Everything is working, however, it is very slow. When I do not use JOIN, it gets really fast, now when I use JOIN, to display 62K is taking more than 2 minutes. This is normal?

mwizeman:
@andrecordeiro I haven't come across this problem yet as my use cases so far have been fairly simple geo shapes. That being said, it's very likely that we'll start looking into geojson and kml files soon so I may have more information for you then. If you find any improvements that need to be made for this use case before we get there, please let us know!

PR #185, opened 2017-11-08
fix version in pom for multi-module test

sfeng88:
We missed the multi-test module poms when releasing vertexium 3.0.1. Might be because the multi-test modules are not enabled by default, they are enabled by a profile.


joeferner:
I just ran through the sonotype deploy instructions as usual which updates all the versions for me. I'm wondering if there is a trick to make it update these versions as well.

PR #191, opened 2017-11-27
fix multimodule test poms

sfeng88:


joeferner:
Grr, I'm going to remember to update this one of these days.

PR #195, opened 2017-11-30
Added: Accumulo in table storage of streaming property value data

joeferner:
Kind of a scary change, not sure if this should hold off until the next major release.

### DIFF
--- a/test/src/main/java/org/vertexium/test/GraphTestBase.java
+++ b/test/src/main/java/org/vertexium/test/GraphTestBase.java
@@ -187,15 +187,18 @@ public abstract class GraphTestBase {
     @Test
     public void testAddStreamingPropertyValue() throws IOException, InterruptedException {
         String expectedLargeValue = IOUtils.toString(new LargeStringInputStream(LARGE_PROPERTY_VALUE_SIZE));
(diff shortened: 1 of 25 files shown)


joeferner:
add unit test to test adding a smaller file after a larger file. This will test if the additional chucks get cleaned up

test/src/main/java/org/vertexium/test/GraphTestBase.java
```diff
@@ -233,11 +236,47 @@ public void testAddStreamingPropertyValue() throws IOException, InterruptedExcep
...
+
+        // now save a smaller value, making sure it gets truncated
+        expectedValue = "small";
+        propLarge = StreamingPropertyValue.create(
+                new ByteArrayInputStream(expectedValue.getBytes()),
+                String.class,
+                expectedValue.length()
+        );
+        graph.prepareVertex("v1", VISIBILITY_A)
+                .addPropertyValue("key1", "largeProp", propLarge, metadata, timestamp + 1, VISIBILITY_A)
```
joeferner: My implementation without the metadata (length field) still worked because I only read the data from the property timestamp. But I added the metadata because it might be useful later.

Super weird but if I take this +1 off of timestamp I end up with duplicate columns (same key) here: https://github.com/visallo/vertexium/pull/195/files#diff-608da76fc5e28276fa9e63d6658f2af7R115 one column with the smaller value and one value with the larger value.

PR #203, opened 2017-12-13
fix pagination returning duplicate ids on different pages

sfeng88:


sfeng88:
@joeferner done

joeferner:
@sfeng88 looks like a legit test failure in testGraphQueryPagingForUniqueIdsSortedOrder for Accumulo

sfeng88:
@joeferner pushed the fix. Couple of things to note. 
1) With this new fix, I made ids sorted ascending order instead of descending order
2) I added a sort by keys(id) to the rows InMemoryTable to match ES and Accumulo

PR #208, opened 2017-12-15
fix pagination returning duplicate ids on different pages

sfeng88:
Let's try this again. Please let travis finish running before merging. Running into weird issues where test pass locally but not in Travis.

NOTE: With this new fix, I made ids sorted ascending order instead of descending order in ES


joeferner:
@sfeng88 you've got conflicts. `CHANGELOG.md` of course

PR #217, opened 2018-01-22
Improve cypher execution memory footprint by using streams where possible in the execution pipeline

joeferner:


cypher/src/main/java/org/vertexium/cypher/ast/model/CypherExpression.java
```diff
@@ -9,6 +9,7 @@
 public abstract class CypherExpression extends CypherAstBase {
     public static String toString(Iterable<CypherAstBase> expressions) {
         return stream(expressions)
+                .map(e -> e == null ? "null" : "")
```
mwizeman: Should this maybe be `.map(e -> e == null ? "null" : e)`?

joeferner: fixed

PR #220, opened 2018-01-25
set size option on es5 scroll api query

sfeng88:


joeferner:
@sfeng88 looks good, just waiting on CI

sfeng88:
Looks like a legitimate test failure. I'll take a look at it

PR #226, opened 2018-02-20
Update Query to Respect FetchHint.INCLUDE_HIDDEN

mwizeman:
Previously, the search index classes were always filtering out results that are hidden. Now, if the FetchHints contain INCLUDE_HIDDEN that filtering will be skipped.

In order to allow this behavior for loadign id's, I also added overloaded versions of `vertexIds`, `edgeIds`, `elementIds`, and `extendedDataRowIds` that take EnumSet<IdFetchHint> indicating whether or not to include hidden.

### DIFF
--- a/core/src/main/java/org/vertexium/query/Query.java
+++ b/core/src/main/java/org/vertexium/query/Query.java
@@ -10,19 +10,75 @@ public interface Query {
 
     QueryResultsIterable<Vertex> vertices(EnumSet<FetchHint> fetchHints);
 
+    /**
+     * Execute the query and return the ids of all matching vertices.
+     * This method will not load the data from storage, which is more
+     * efficient in cases where you only need to know the ids of matching vertices.
+     * Hidden vertices are not included in the results.
+     *
+     * @return The ids of vertices that match this query.
+     */
     QueryResultsIterable<String> vertexIds();
+
+    /**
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/vertexium/query/Query.java
```diff
@@ -10,20 +10,76 @@
...
 
+    /**
+     * Execute the query and return the ids of all matching vertices.
+     * This method will not load the data from storage, which is more
+     * efficient in cases where you only need to know the ids of matching vertices.
+     *
+     * @param includeHidden Whether or not to include hidden vertices in the results
+     * @return The ids of vertices that match this query.
+     */
+    QueryResultsIterable<String> vertexIds(boolean includeHidden);
```
joeferner: I don't love the mystery boolean parameter. It feels a little weird to pass `FetchHint`s here though. Maybe `IdFetchHint`s which could be an `EnumSet` or just an `enum` with two values? Not really sure.

joeferner:
Also can you rebase and add to the changelog

mwizeman:
Some tests are failing. I'll fix that and address the PR comments.

mwizeman:
@joeferner Fixed. I went with `EnumSet<IdFetchHint>` to make it match our use of EnumSet<FetchHint>. It also means that we can add more hints in the future without changing the API methods.

PR #227, opened 2018-02-21
fix wrong vertex being returned after readding the same vertex with a different visibility after soft delete

sfeng88:


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # v3.1.2
+* Fixed: InMemory and Accumulo fix wrong vertex being returned after readding the same vertex with a different visibility after soft delete
```
mwizeman: typo

accumulo-iterators/src/main/java/org/vertexium/accumulo/iterator/ElementIterator.java
```diff
@@ -112,10 +112,12 @@ private boolean processKeyValue(Key key, Text columnFamily, Value value) {
         }
 
         if (getVisibilitySignal().equals(columnFamily)) {
-            elementData.visibility = key.getColumnVisibility();
-            elementData.timestamp = key.getTimestamp();
-            processSignalColumn(key.getColumnQualifier());
-            return true;
+            if (elementData.timestamp == 0 || key.getTimestamp() > elementData.timestamp) {
```
joeferner: When is timestamp 0?

sfeng88: `elementData.timestamp` starts off as 0 before it is set the first time. We probably don't need that extra check since the second half of the conditional will catch it. I'll remove it and push the change.

sfeng88: I cleaned up the if statement

Issue #178, opened 2017-10-31
`markPropertyHidden` does not Update the Search Index


mwizeman:
When `markPropertyHidden` and `markPropertyVisible` are called on the `AccumuloGraph`, the search index is not updated to reflect it.

mwizeman:
Once this issue is fixed, [re-evaluate the changes made in this PR](https://github.com/visallo/vertexium/pull/235) to see if they are still necessary.

PR #235, opened 2018-02-27
Query.hasAuthorization match elements with hidden properties

mwizeman:
If the only use of an authorization was a hidden property, `Query.hasAuthorization` was not returning an element since the search index was not being updated for hidden properties. This PR makes such elements return for `Query.hasAuthorization`, but [does not fix the issue of the hidden properties still being searchable](https://github.com/visallo/vertexium/issues/178). Once that issue is fixed, the `HIDDEN_PROPERTY_FIELD_NAME` can likely be removed.

Known Issue: If a property is marked hidden, it can still be found with a search but it will not be retrieved from the data store.

### DIFF
--- a/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
+++ b/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
@@ -78,6 +78,7 @@ public class Elasticsearch5SearchIndex implements SearchIndex, SearchIndexWithVe
     public static final String ELEMENT_TYPE_FIELD_NAME = "__elementType";
     public static final String VISIBILITY_FIELD_NAME = "__visibility";
     public static final String HIDDEN_VERTEX_FIELD_NAME = "__hidden";
+    public static final String HIDDEN_PROPERTY_FIELD_NAME = "__hidden_property";
     public static final String OUT_VERTEX_ID_FIELD_NAME = "__outVertexId";
     public static final String IN_VERTEX_ID_FIELD_NAME = "__inVertexId";
     public static final String EDGE_LABEL_FIELD_NAME = "__edgeLabel";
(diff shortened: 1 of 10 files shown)


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -78,6 +78,7 @@
     public static final String ELEMENT_TYPE_FIELD_NAME = "__elementType";
     public static final String VISIBILITY_FIELD_NAME = "__visibility";
     public static final String HIDDEN_VERTEX_FIELD_NAME = "__hidden";
+    public static final String HIDDEN_PROPERTY_FIELD_NAME = "__hidden_property";
```
joeferner: I'm guessing we don't need to add it here because it will always have visibility correct? https://github.com/visallo/vertexium/blob/a384cebded60878f47d3e00084473e7d255616e1/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java#L1442

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -755,6 +756,38 @@ public void markElementVisible(Graph graph, Element element, Visibility visibili
...
+                String indexName = getIndexName(element);
+                IndexInfo indexInfo = ensureIndexCreatedAndInitialized(graph, indexName);
+                addPropertyToIndex(graph, indexInfo, hiddenVisibilityPropertyName, visibility, Boolean.class, false, false, false);
+            }
+
+            XContentBuilder jsonBuilder = XContentFactory.jsonBuilder().startObject();
+            jsonBuilder.field(hiddenVisibilityPropertyName, true);
+            jsonBuilder.endObject();
+
+            getClient()
```
joeferner: Depending on who merges first we'll need to add the push change on refresh

PR #241, opened 2018-03-05
Prevent InMemoryGraph Concurrent Modification Exceptions

mwizeman:
Synchronize the methods and also returning Collection copies.


mwizeman:
Oops. Looks like we already push this change on Friday.

PR #243, opened 2018-03-05
Apply ExistingElementMutations with Painless Script in Elasticsearch

mwizeman:
When saving an ExistingElementMutation, the AccumuloElement/AccumuloGraph code would previously make multiple index update requests and up to two refreshes of the search index. This change adds a new method to the SearchIndex to apply an ExistingElementMutation directly. This allows the ES index to be updated with a single request using a painless script. 

As an additional benefit, we were previously sending the entire document as an upsert request which potentially sent more data than necessary. Now we only send the property value for fields that are added by the mutation.

### DIFF
--- a/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
+++ b/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
@@ -367,7 +367,7 @@ public class Elasticsearch5SearchIndex implements SearchIndex, SearchIndexWithVe
 
     @Override
     public <TElement extends Element> void updateElement(Graph graph, ExistingElementMutation<TElement> mutation, Authorizations authorizations) {
-        Element element = mutation.getElement();
+        TElement element = mutation.getElement();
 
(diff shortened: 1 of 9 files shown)


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -368,6 +365,75 @@ private void addElementWithScript(Graph graph, Element element, Authorizations a
         }
     }
 
+    @Override
+    public <TElement extends Element> void updateElement(Graph graph, ExistingElementMutation<TElement> mutation, Authorizations authorizations) {
```
joeferner: To make this method a little more readable/shorter can you split it up into 3-4 steps perhaps

mwizeman: @joeferner Good idea. Done with https://github.com/visallo/vertexium/pull/243/commits/6359cf5913d91630bebb0121a9dcb9b3ee65b8e3

PR #255, opened 2018-03-28
Added: Method to Vertex to get more detailed edge summary

joeferner:
Can you think of anything else we need to get inverse of relationship details pane working?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * Changed: FetchHints to support more filtering
 * Changed: Accumulo: Store metadata in an indexed list to prevent duplication in memory and over the wire 
 * Fixed: Accumulo Iterator memory leak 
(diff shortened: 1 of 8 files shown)


joeferner:
Should it be `EdgesSummary` or `EdgeSummary`.

Should we deprecate
- https://github.com/visallo/vertexium/blob/master/core/src/main/java/org/vertexium/Vertex.java#L214
- https://github.com/visallo/vertexium/blob/master/core/src/main/java/org/vertexium/Vertex.java#L223

mwizeman:
I think EdgesSummary is fine. 

I can go either way on deprecating those two methods. If we keep them, should they just be wrappers around the new `getEdgesSummary` method? Also, I noticed that this `getEdgeCount` method warns if the FetchHints aren't right but the others don't. Should they?

PR #259, opened 2018-04-03
Move property conversion so it's called on updates

jharwig:
If calling index.updateElement the conversion wasn't called so would fail if saving a BigDecimal or BigInteger

All tests use InMemoryElement which always calls `addElement`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 # v3.2.2
 * Fixed: Improved support for multithreaded clients with InMemoryGraph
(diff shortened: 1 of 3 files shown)


jharwig:
@joeferner  Should we change InMemoryElement to call updateElement for updates too?

joeferner:
I'm a little surprised this isn't caught since the multi-module test will run Accumulo and ES plugins together. I think there are code paths we just must not be testing.

sfeng88:
@jharwig can you add a changelog entry for this fix please?

jharwig:
@sfeng88 done, also found a test that breaks before the fix. Needed to use `prepareMutation` to update an element not `prepareUpdate`

PR #272, opened 2018-04-30
Multiple calls to Query.hasId Should AND rather than OR

mwizeman:


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
```diff
@@ -232,12 +232,12 @@ protected QueryBuilder createQueryStringQuery(QueryStringQueryParameters queryPa
         }
         if ((elementTypes == null || elementTypes.contains(ElasticsearchDocumentType.EDGE))
                 && getParameters().getEdgeLabels().size() > 0) {
-            String[] edgeLabelsArray = getParameters().getEdgeLabels().toArray(new String[getParameters().getEdgeLabels().size()]);
+            String[] edgeLabelsArray = getParameters().getEdgeLabels().toArray(new String[0]);
```
joeferner: Any reason not to pass the size here? It's a small optimization but it does reduce the number of created objects by 1.

mwizeman: I thought that too, but an IntelliJ inspection suggested we change that. It said:

>There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
>
>In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.

I did some quick looking and found that people had benchmarked and their results seem to agree: https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion

I'm happy to switch it back if you like.

joeferner: Nope, didn't realize this changed. I've just been doing it so long I just assumed using presized arrays was still faster

PR #278, opened 2018-05-18
Give FetchHintsBuilder ability to parse json

joeybrk372:


core/src/main/java/org/vertexium/FetchHintsBuilder.java
```diff
@@ -55,6 +59,44 @@ public FetchHints build() {
...
+        if (jsonArray == null) {
+            return null;
+        }
+
+        List<String> list = new ArrayList<>();
+        for (int i = 0;i < jsonArray.length(); i++) {
+            list.add((String) jsonArray.get(i));
+        }
+
+        return ImmutableSet.<String>builder().addAll(list).build();
```
joeferner: Don't make this an `ImmutableSet` when the builder gets converted to a `FetchHints` it will take care of that

core/src/main/java/org/vertexium/FetchHintsBuilder.java
```diff
@@ -55,6 +59,44 @@ public FetchHints build() {
         );
     }
 
+    public FetchHintsBuilder parse(JSONObject fetchHintsJson) {
```
joeferner: I'm feeling like this should be a a static method which returns a new `FetchHintsBuilder`. It would be weird to create a builder, set some properties then call parse and it would wipe everything out.

PR #281, opened 2018-05-19
Added: Option to Elasticsearch index to log large document (search.logRequestSizeLimit)

joeferner:


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -1679,11 +1682,32 @@ public void addElements(Graph graph, Iterable<? extends Element> elements, Autho
...
+        }
+        int sizeInBytes = 0;
+        if (request.doc() != null) {
+            sizeInBytes += request.doc().source().length();
+        }
+        if (request.upsertRequest() != null) {
+            sizeInBytes += request.upsertRequest().source().length();
+        }
+        if (request.script() != null) {
+            sizeInBytes += request.script().getIdOrCode().length() * 2;
```
mwizeman: Why `* 2`?

joeferner: Beats me. I should probably add a comment here that we found the code here: https://github.com/elastic/elasticsearch/blob/916bf9d26d6292dcb9b229123c58a86f0e741ffd/server/src/main/java/org/elasticsearch/action/bulk/BulkRequest.java#L180

mwizeman: Works for me.

PR #286, opened 2018-06-14
Per query scoring strategy

joeferner:
For fun, I wanted to add support for hamming distance scoring to play with a couple algorithms. This PR deprecates the old unused global ES scoring strategy and replaces it with a per query scoring strategy. One of which is hamming distance.


elasticsearch5/src/test/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndexTest.java
```diff
@@ -175,4 +176,9 @@ private Throwable getRootCause(Throwable e) {
         }
         return getRootCause(e.getCause());
     }
+
+    @Override
+    protected ScoringStrategy getHammingDistanceScoringStrategy(String field, String hash) {
+        return new ElasticsearchHammingDistanceScoringStrategy(field, hash);
```
joeferner: I would love to avoid having to create scoring strategies specific to ES but I can't think of a way around that.

Issue #295, opened 2018-07-26
Query for edges with vertex id on either side of the edge


joeferner:
```
QueryResultsIterable<Edge> edgeInVertexIdMatch = graph.query(authorizations)
               .has(Edge.VERTEX_IDS_PROPERTY_NAME, Contains.IN, vertexIds)
               .edges();
```

PR #296, opened 2018-07-26
Query edges with a given in or out vertex

joeferner:
completes... https://github.com/visallo/vertexium/issues/295

@jharwig can you look at the unit test in `GraphTestBase` and make sure this will do what you need.

### DIFF
--- a/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
+++ b/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
@@ -276,6 +276,8 @@ public class ElasticsearchSearchQueryBase extends QueryBase {
                                 .unmappedType(KEYWORD_UNMAPPED_TYPE)
                                 .order(esOrder)
                 );
+            } else if (Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME.equals(sortContainer.propertyName)) {
+                throw new VertexiumException("Cannot sort by " + Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME);
             } else if (ExtendedDataRow.TABLE_NAME.equals(sortContainer.propertyName)) {
                 q.addSort(
                         SortBuilders.fieldSort(Elasticsearch5SearchIndex.EXTENDED_DATA_TABLE_NAME_FIELD_NAME)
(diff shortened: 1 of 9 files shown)


jharwig:
@sfeng88 Wrote that code, so she would be best to answer that

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
```diff
@@ -998,64 +1006,67 @@ protected QueryBuilder getFilterForComparePredicate(Compare compare, HasValueCon
...
+        } else if (Edge.LABEL_PROPERTY_NAME.equals(propertyName)) {
+            propertyName = Elasticsearch5SearchIndex.EDGE_LABEL_FIELD_NAME;
+        } else if (Edge.OUT_VERTEX_ID_PROPERTY_NAME.equals(propertyName)) {
+            propertyName = Elasticsearch5SearchIndex.OUT_VERTEX_ID_FIELD_NAME;
+        } else if (Edge.IN_VERTEX_ID_PROPERTY_NAME.equals(propertyName)) {
+            propertyName = Elasticsearch5SearchIndex.IN_VERTEX_ID_FIELD_NAME;
+        } else if (Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME.equals(propertyName)) {
+            return QueryBuilders.boolQuery()
+                    .should(getFilterForProperty(compare, has, Edge.OUT_VERTEX_ID_PROPERTY_NAME, value))
+                    .should(getFilterForProperty(compare, has, Edge.IN_VERTEX_ID_PROPERTY_NAME, value));
```
mwizeman: Since this is an `or` with possibly other criteria involved, should we be safe and set the `minimum_should_match` to `1`? This is from the docs for `should`:
```
If the bool query is in a query context and has a must or 
filter clause then a document will match the bool query even
if none of the should queries match.  In this case these 
clauses are only used to influence the score.
...
This behavior may be explicitly controlled by settings the minimum_should_match parameter.
```

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
```diff
@@ -954,34 +956,40 @@ protected QueryBuilder getFilterForContainsPredicate(Contains contains, HasValue
...
+        } else if (Edge.LABEL_PROPERTY_NAME.equals(propertyName)) {
+            propertyName = Elasticsearch5SearchIndex.EDGE_LABEL_FIELD_NAME;
+        } else if (Edge.OUT_VERTEX_ID_PROPERTY_NAME.equals(propertyName)) {
+            propertyName = Elasticsearch5SearchIndex.OUT_VERTEX_ID_FIELD_NAME;
+        } else if (Edge.IN_VERTEX_ID_PROPERTY_NAME.equals(propertyName)) {
+            propertyName = Elasticsearch5SearchIndex.IN_VERTEX_ID_FIELD_NAME;
+        } else if (Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME.equals(propertyName)) {
+            return QueryBuilders.boolQuery()
+                    .should(getFilterForProperty(contains, has, Edge.OUT_VERTEX_ID_PROPERTY_NAME, value))
+                    .should(getFilterForProperty(contains, has, Edge.IN_VERTEX_ID_PROPERTY_NAME, value));
```
mwizeman: Since this is an `or` with possibly other criteria involved, should we be safe and set the `minimum_should_match` to `1`? This is from the docs for `should`:
```
If the bool query is in a query context and has a must or 
filter clause then a document will match the bool query even
if none of the should queries match.  In this case these 
clauses are only used to influence the score.
...
This behavior may be explicitly controlled by settings the minimum_should_match parameter.
```

joeferner:
@mwizeman added min match == 1.

PR #302, opened 2018-08-29
Return ElasticsearchEdge and ElasticsearchVertex when FetchHints.EDGE_REF and FetchHints.NONE is set for performance

sfeng88:


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/ElasticsearchEdge.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public String getVertexId(Direction direction) {
+        if (direction.equals(Direction.IN)) {
+            return inVertexId;
+        } else if (direction.equals(Direction.OUT)) {
+            return outVertexId;
+        }
+        return null;
```
mwizeman: Should we throw a VertexiumNotSupportedException exception here? I think I'd expect it to blow up if I passed in `Direction.BOTH`

PR #311, opened 2018-09-24
Add Compare.STARTS_WITH Predicate

mwizeman:
Added the ability to perform a term query by prefix using a new Compare.STARTS_WITH predicate.


mwizeman:
@joeferner Made a small tweak to allow the prefix query to be used when searching GeoShapes as well as Text.

PR #315, opened 2018-10-07
Elasticsearch: support query from ElasticsearchVertex

joeferner:
I can't remember what I needed this for, I can look Tuesday but I know this fixed my issue


mwizeman:
IIRC, there's a bit of ingest code loading vertices with `FetchHints.NONE` and then querying their edges. That should work out since all you need to query the edges is the original vertex ID.

PR #316, opened 2018-10-07
Changed: Accumulo to lazily load metadata values

joeferner:
I'm not 100% sure how this is going to effect serialization. I'm pretty sure we don't serialize metadata in the Accumulo graph anywhere but the metadata entry class is marked serializable.


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # v4.3.2
-* Added: The ability to perform a term query by prefix using a new Compare.STARTS_WITH predicate. 
+* Added: The ability to perform a term query by prefix using a new Compare.STARTS_WITH predicate.
+* Changed: Accumulo to lazily load metadata values 
 * Fixed: AccumuloGraphLogger was throwing an error when trying to log a field from a Scanner that was removed in Accumulo 1.8.0
```
mwizeman: Should you mention that you converted Metadata to an Interface? Code will need to be updated to `Metadata.create` instead of `new Metadata`.

accumulo/src/main/java/org/vertexium/accumulo/LazyPropertyMetadata.java
```diff
@@ -1,14 +1,321 @@
...
+            }
+            metadataIndexes = new int[0];
+            removedEntries.clear();
+        } finally {
+            getEntriesLock().writeLock().unlock();
+        }
+    }
+
+    private void loadAll() {
+        if (metadataEntries != null && metadataIndexes != null) {
```
mwizeman: Should this method acquire the read lock?

PR #320, opened 2018-10-26
Invalidate the define property cache (4.1.1.x)

jharwig:
When we get a metadata change event that is from a define property change, also invalidate the cache in GraphBase.

Fixes issues where multiple graph processes can start seeing elements and properties that don't exist in the graphs local caches.


jharwig:
I could create a test that invalidation works. Probably harder to check the multiple graphs with curator syncing?

jharwig:
Trying to reproduce a race condition locally that happened in code build…

jharwig:
@joeferner @mwizeman Added unit test and changed the way I did it a little to fix some race conditions. I didn't realize the curator handler was fired for the current instance. Caused it to invalidate the whole cache on every property definition change. I changed to only invalidate the one value.

PR #329, opened 2018-11-17
EdgeInfo.getDirection was reversed when loaded from InMemoryVertex

mwizeman:


mwizeman:
Looks like the Cypher tests are failing with this change. I'll look into that.

mwizeman:
Cypher is fixed.

PR #335, opened 2018-12-21
Fix Extended Data Query with no Access to Columns

mwizeman:
Updated to make sure that you have access to at least one column for a query to match extended data. Previously, if you had access to the element but none of the extended data columns you could still match on the row using a query with no filters. This resulted in the row being counted in `totalHits` but not loaded by Accumulo, which caused errors in the paging iterables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-# v4.4.2
-* Fixed: If you had access an element but none of the extended data columns you could still match on the extended data row using a query with no filters. This would result in the row being counted in `totalHits`, but it could not be retrieved from Accumulo.
+# v4.5.0
+* Fixed: If you had access an element but none of the extended data columns you could still match on the extended data row using a query with no filters. This would result in the row being counted in `totalHits`, but it could not be retrieved from Accumulo. NOTE: This change requires that all extended data values are re-indexed in ElasticSearch.
 
 # v4.4.2
 * Fixed: `EdgeInfo.getDirection` was reversed when loaded from InMemoryVertex.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # v4.4.2
+* Fixed: If you had access an element but none of the extended data columns you could still match on the extended data row using a query with no filters. This would result in the row being counted in `totalHits`, but it could not be retrieved from Accumulo.
+
+# v4.4.2
 * Fixed: `EdgeInfo.getDirection` was reversed when loaded from InMemoryVertex.
 * Changed: Elasticsearch: index on reads instead of indexing on writes.
 * Changed: Elasticsearch: optimize adding properties to vertices/edges with lots of properties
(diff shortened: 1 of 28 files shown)


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -137,6 +140,26 @@ public Elasticsearch5SearchIndex(Graph graph, GraphConfiguration config) {
         this.logRequestSizeLimit = this.config.getLogRequestSizeLimit();
         this.exceptionHandler = this.config.getExceptionHandler(graph);
         this.flushObjectQueue = new FlushObjectQueue(this);
+
+        storePainlessScript("addFieldsToExtendedDataScript", getClass().getResourceAsStream("add-fields-to-extended-data.painless"), true);
```
joeferner: Move `getClass().getResourceAsStream` into `storePainlessScript`.

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -137,6 +140,26 @@ public Elasticsearch5SearchIndex(Graph graph, GraphConfiguration config) {
...
+        storePainlessScript("updateFieldsOnDocumentScript", getClass().getResourceAsStream("update-fields-on-document.painless"), false);
+    }
+
+    private void storePainlessScript(String scriptId, InputStream scriptSource, boolean includeHelpers) {
+        try {
+            String source = IOUtils.toString(scriptSource);
+            if (includeHelpers) {
+                source = IOUtils.toString(getClass().getResourceAsStream("helper-functions.painless")) + " " + source;
+            }
+            source = source.replaceAll("\\r?\\n", " ");
```
joeferner: replace all `"` with `\"`

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.4.2
+* Fixed: If you had access an element but none of the extended data columns you could still match on the extended data row using a query with no filters. This would result in the row being counted in `totalHits`, but it could not be retrieved from Accumulo.
```
joeferner: Does this require a reindex? If so lets bump it to at least 4.5

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
mwizeman: Fixed with: 575b8a8b960d6b77f8828e2c85bf1347503651c6

elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
mwizeman: Fixed with: 575b8a8b960d6b77f8828e2c85bf1347503651c6

CHANGELOG.md
mwizeman: It does. Fixed with: 575b8a8b960d6b77f8828e2c85bf1347503651c6

PR #336, opened 2019-01-02
Fix __elementType Visibility on Extended Data

mwizeman:
When the visibility changes on an element, also update the visibility of the `__elementType` field on all associated extended data.

### DIFF
--- a/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
+++ b/elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
@@ -4,7 +4,9 @@ import com.carrotsearch.hppc.cursors.ObjectCursor;
 import com.google.common.base.Joiner;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse;
 import org.elasticsearch.action.admin.cluster.node.info.NodeInfo;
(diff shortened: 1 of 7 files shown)


elasticsearch5/src/main/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndex.java
```diff
@@ -456,6 +455,21 @@ public void addElement(Graph graph, Element element, Authorizations authorizatio
             getIndexRefreshTracker().pushChange(indexInfo.getIndexName());
             addActionRequestBuilderForFlush(element, updateRequestBuilder);
 
+            ImmutableSet<String> extendedDataTableNames = mutation.getElement().getExtendedDataTableNames();
```
joeferner: What happens when fetch hints don't include extended data table names?

mwizeman: Then the extended data will still have the old visibility on them. That's basically the behavior that you get today without this PR, but all the time. I considered warning if there is no extended data but there's no way to tell if there legitimately isn't extended data or if it just wasn't loaded due to visibilities or fetch hints.

This PR makes it better, but it's still pretty possible that things can get out of sync.

joeferner: I thought getExtendedDataTableNames threw an exception of they weren't fetched. If it doesn't it probably should to be consistent.

mwizeman: It looks like it simply returns the set that was passed in when the element object was created: https://github.com/visallo/vertexium/blob/master/accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java#L386-L388

If we do change that to throw an exception, how would this new code tell whether there was something to do without encountering an exception? I guess it could first check the fetch hints and then try to get the table names. Shall I try that?

joeferner: I think that would be the most consistent way of doing things. I think we throw exceptions everywhere else if the fetch hints are missing.

PR #339, opened 2019-01-14
Add geopoint accuracy

joeferner:
This is the first step which does not require re-indexing geopoint objects. The next step will be to re-index geopoint objects to geo-circles.


core/src/main/java/org/vertexium/type/GeoPoint.java
```diff
@@ -92,6 +118,7 @@ public int hashCode() {
         hash = 47 * hash + (int) (Double.doubleToLongBits(this.latitude) ^ (Double.doubleToLongBits(this.latitude) >>> 32));
         hash = 47 * hash + (int) (Double.doubleToLongBits(this.longitude) ^ (Double.doubleToLongBits(this.longitude) >>> 32));
         hash = 47 * hash + (this.altitude != null ? this.altitude.hashCode() : 0);
+        hash = this.accuracy != null ? (47 * hash + this.accuracy.hashCode()) : 0;
```
mwizeman: This looks like a typo. If no accuracy is provided, the hash will always be 0

joeferner: Wow, how did you catch that?

mwizeman: Reading through the changes, it caught my attention that it didn't look like the other lines above it :-)

joeferner:
This PR should be good to go now

PR #342, opened 2019-02-05
Cypher: update to latest openCypher tck

joeferner:
Thought I might update the Cypher TCK. Ended up being a bit more work that expected. Big thing to note is that no additional tests are broken but we are missing some new functionality.


mwizeman:
The `oc` prefix everywhere is a little annoying, but I like that handling some of the operators got better. You were able to use methods rather than looking for magic strings in many places.

joeferner:
Yeah the OC thing is annoying not sure if you can get antlr too remove
those. I get the Grammer file directly from them.

PR #343, opened 2019-02-19
add sync index

MrDLontheway:
When we write bulk data, if the index is written to accumulo and es one by one, we can first batch write accumulo in asynchronous synchronous es index data, which can greatly improve the efficiency of batch import data.
  After the test index synchronization speed is greatly improved compared to before


joeferner:
Thanks for the contribution. I think we have a call already that does what you want `Graph.saveElementMutations` https://github.com/visallo/vertexium/blob/master/core/src/main/java/org/vertexium/Graph.java#L1080 and the implementation is similar to what you provided https://github.com/visallo/vertexium/blob/master/core/src/main/java/org/vertexium/GraphBaseWithSearchIndex.java#L170

If I'm wrong feel free to reopen this PR.

PR #350, opened 2019-03-20
add es not index  property

MrDLontheway:


joeferner:
Looks like there are test failures, can you take a look at those.

PR #351, opened 2019-04-24
Deduplication iterator

dspoja:
Here is the row deduplication iterator. We should plan on testing it with real data too.


sfeng88:
@dspoja There are test failures, can you take a look at this PR?

dspoja:
The test failures are occurring because the Travis job times out while waiting for the unit tests to complete. Before these changes, accumulo test were taking about 16 minutes to complete, and the entire test suite taking about 40 minutes. Now, with the addition of the FlushEvent and that fact that we call flush a lot in the unit tests and the flush events invoke the new iterator which attempts to deduplicate rows on each flush, the accumulo tests take about 1 hour to run. In my local environment, all tests complete successfully, but they take over 1 hour to run. Commenting out `queueEvent()` call on line 1479 in AccumuloGraph.java shortens the test time back to 16 minutes, but then the new iterator is not run. Extending the Travis timeout would allow the build to succeed. This PR needs to be tested with real data before it is merged.

sfeng88:
@dspoja Is it possible to split the test that is taking a really long time to run into it's own profile? That way we can get travis to run it in parallel as well

dspoja:
@sfeng88 I don't think that is possible, since it is basically every test in GraphBaseTest that has a flush event, so each test takes a little longer to complete, and they all together add up to significant delay. If there were only a couple of tests that were taking too long, we could definitely separate them.

sfeng88:
@mwizeman @joeferner Thoughts about this PR?

accumulo-iterators/pom.xml
```diff
@@ -27,7 +27,8 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <version>4.12</version>
```
joeferner: This version should be coming from the root directory pom

accumulo-iterators/src/main/java/org/vertexium/accumulo/iterator/RowDeduplicationIterator.java
```diff
@@ -0,0 +1,218 @@
...
+    private static final long DEFAULT_MAX_BUFFER_SIZE = Long.MAX_VALUE;
+    private static final List<String> COLUMN_FAMILY_TYPES = Arrays.asList("PROP", "PROPD");
+    protected SortedKeyValueIterator<Key, Value> sourceIter;
+    private Key topKey = null;
+    private Value topValue = null;
+    private long maxBufferSize = DEFAULT_MAX_BUFFER_SIZE;
+
+    @Override
+    public SortedKeyValueIterator<Key, Value> deepCopy(IteratorEnvironment env) {
+        RowDeduplicationIterator newInstance;
```
joeferner: probably should just `new RowDeduplicationIterator()` here. Using reflection here probably not needed.

accumulo-iterators/src/main/java/org/vertexium/accumulo/iterator/RowDeduplicationIterator.java
```diff
@@ -0,0 +1,218 @@
...
+            newInstance = this.getClass().newInstance();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        newInstance.sourceIter = sourceIter.deepCopy(env);
+        newInstance.maxBufferSize = maxBufferSize;
+        return newInstance;
+    }
+
+    List<KeyValue> sortedRow = new ArrayList<>();
```
joeferner: move variables to the top. Mark them as private.

accumulo-iterators/src/test/java/org/vertexium/accumulo/iterator/RowDeduplicationIteratorTest.java
```diff
@@ -0,0 +1,204 @@
...
+import org.apache.accumulo.core.iterators.IteratorUtil;
+import org.apache.accumulo.core.iterators.SortedMapIterator;
+import org.apache.commons.collections.BufferOverflowException;
+import org.apache.hadoop.io.Text;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.*;
+
+public class RowDeduplicationIteratorTest {
```
joeferner: Can you test interleaved properties.

In time order...
- t1 - prop1: John
- t2 - prop2: Doe
- t3 - prop1: John
- t4 - prop2: Smith

The two `John` values should be collapsed.

PR #360, opened 2019-05-01
Accumulo use a thread pool when doing batch scans

joeferner:
Benchmark https://github.com/visallo/vertexium/compare/fast-batch-reader?expand=1#diff-7de8e1270d184f1d847439d3722b3a36R9015

Before change: 36.852s
After change: 24.665s

This change also alleviates the strict need to close the iterator.

### DIFF
--- a/accumulo/src/main/java/org/vertexium/accumulo/util/ConnectorUtils.java
+++ b/accumulo/src/main/java/org/vertexium/accumulo/util/ConnectorUtils.java
@@ -1,0 +1,23 @@
+package org.vertexium.accumulo.util;
+
+import org.apache.accumulo.core.client.Connector;
+import org.apache.accumulo.core.client.impl.ClientContext;
+import org.apache.accumulo.core.client.impl.ConnectorImpl;
+import org.vertexium.VertexiumException;
+
+import java.lang.reflect.Field;
+
+public class ConnectorUtils {
+    public static ClientContext getContext(Connector connector) {
+        if (!(connector instanceof ConnectorImpl)) {
+            throw new VertexiumException("Invalid connector type. Expected " + ConnectorImpl.class.getName() + " found " + connector.getClass().getName());
+        }
+        try {
+            Field f = ConnectorImpl.class.getDeclaredField("context");
+            f.setAccessible(true);
+            return (ClientContext) f.get(connector);
+        } catch (Exception e) {
+            throw new VertexiumException("Could not get context field", e);
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v4.6.0
+* Changed: Accumulo use a thread pool when doing batch scans 
+
 # v4.5.2
 * Fixed: Accumulo delete extended data not removing rows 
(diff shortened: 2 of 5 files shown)


accumulo/src/main/java/org/vertexium/accumulo/util/ConnectorUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+public class ConnectorUtils {
+    public static ClientContext getContext(Connector connector) {
+        if (!(connector instanceof ConnectorImpl)) {
+            throw new VertexiumException("Invalid connector type. Expected " + ConnectorImpl.class.getName() + " found " + connector.getClass().getName());
+        }
+        try {
+            Field f = ConnectorImpl.class.getDeclaredField("context");
```
mwizeman: This won't be necessary when we upgrade to Accumulo 2.0. They seem to have added a get method: https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/clientImpl/ConnectorImpl.java#L74

accumulo/src/main/java/org/vertexium/accumulo/util/ConnectorUtils.java
```diff
@@ -0,0 +1,38 @@
...
+            throw new VertexiumException("Could not get context field", e);
+        }
+    }
+
+    public static String getTableId(Connector connector, String tableName) throws TableNotFoundException {
+        if (!(connector instanceof ConnectorImpl)) {
+            throw new VertexiumException("Invalid connector type. Expected " + ConnectorImpl.class.getName() + " found " + connector.getClass().getName());
+        }
+        try {
+            Method m = ConnectorImpl.class.getDeclaredMethod("getTableId", String.class);
```
mwizeman: Looks like you might be able to use this helper instead: https://github.com/apache/accumulo/blob/rel/1.9.3/core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java#L68

joeferner: I saw that, it's takes an `instance` which we don't keep around. Looks like in Accumulo 2.0 we can get that too... https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/clientImpl/ConnectorImpl.java#L96 (context comes from `getAccumuloClient()`)

accumulo/src/main/java/org/vertexium/accumulo/util/ConnectorUtils.java
joeferner: Excellent

accumulo/src/main/java/org/vertexium/accumulo/util/ConnectorUtils.java
mwizeman: I think you should be able to ask the `Connector` for its `Instance`: https://github.com/apache/accumulo/blob/rel/1.9.3/core/src/main/java/org/apache/accumulo/core/client/Connector.java#L216

joeferner: done

PR #369, opened 2019-06-05
Refactor ElementId to allow Elements to comply with the interface

joeferner:
This allows methods to take an `ElementId` as a parameter and you could just pass the `Element` itself.

FYI makes some of my merge code easier


joeferner:
@mwizeman  We should probably come up with a new name for `ElementId` and `ElementLocation`

joeferner:
OK, I think I got this one cleaned up

PR #378, opened 2019-07-08
Fixed: Query on Extended Data reserved field names (ELEMENT_TYPE, ELEMENT_ID, TABLE_NAME, ROW_ID)

joeferner:


core/src/main/java/org/vertexium/ElementBase.java
```diff
@@ -45,25 +65,10 @@ public Property getProperty(String name, Visibility visibility) {
...
-        } else if (Edge.OUT_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge) {
-            ArrayList<Property> result = new ArrayList<>();
-            result.add(getOutVertexIdProperty());
-            return result;
-        } else if (Edge.IN_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge) {
-            ArrayList<Property> result = new ArrayList<>();
-            result.add(getInVertexIdProperty());
-            return result;
-        } else if (Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge) {
+        Property reservedProperty = getReservedProperty(name);
```
mwizeman: Do you need to do this? It looks like the implementation of this method just be `return getProperties(null, name)`?

accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java
```diff
@@ -230,16 +230,9 @@ public Property getProperty(String key, String name) {
...
-            return getIdProperty();
-        } else if (Edge.LABEL_PROPERTY_NAME.equals(name) && this instanceof Edge) {
-            return getEdgeLabelProperty();
-        } else if (Edge.OUT_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge) {
-            return getOutVertexIdProperty();
-        } else if (Edge.IN_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge) {
-            return getInVertexIdProperty();
-        } else if (Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge) {
-            return getInOrOutVertexIdProperty();
+        Property reservedProperty = getReservedProperty(name);
```
mwizeman: This check for reservedProperty will be taken care of in the call to `getProperties(name)` below.

accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java
```diff
@@ -250,16 +243,9 @@ public Property getProperty(String name) {
...
+        if (reservedProperty != null) {
+            return reservedProperty.getValue();
         }
         getFetchHints().assertPropertyIncluded(name);
```
mwizeman: Move this check into `getPropertyValue(String key, String name, int index)`, gut this method and have it just call the other one.

accumulo/src/main/java/org/vertexium/accumulo/AccumuloElement.java
```diff
@@ -330,7 +309,11 @@ protected void setVisibility(Visibility visibility) {
             || (Edge.LABEL_PROPERTY_NAME.equals(name) && this instanceof Edge)
             || (Edge.OUT_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge)
             || (Edge.IN_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge)
-            || (Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge)) {
+            || (Edge.IN_OR_OUT_VERTEX_ID_PROPERTY_NAME.equals(name) && this instanceof Edge)
+            || (ExtendedDataRow.TABLE_NAME.equals(name) && this instanceof ExtendedDataRow)
+            || (ExtendedDataRow.ELEMENT_TYPE.equals(name) && this instanceof ExtendedDataRow)
+            || (ExtendedDataRow.ELEMENT_ID.equals(name) && this instanceof ExtendedDataRow)
+            || (ExtendedDataRow.ROW_ID.equals(name) && this instanceof ExtendedDataRow)) {
```
mwizeman: In order to prevent having to repeat these property names, could this whole `if` be replaced with?
```
Property reservedProperty = getReservedProperty(name);
if (reservedProperty != null) {
    ArrayList<Property> result = new ArrayList<>();
    result.add(reservedProperty);
    return result;
}
```

PR #381, opened 2019-07-22
Fixed: Elasticsearch Compare.NOT_EQUAL and Contains.NOT_IN with multiple visibilities

joeferner:


mwizeman:
Waiting for Travis to pass... The Accumulo tests had an error but I suspect it's a false alarm. I restarted it.

PR #400, opened 2019-09-26
Only wait for items added before flush was called to be flushed

joeferner:
Before this, one thread could cause another thread to wait indefinitly as
long as new writes are coming in.


joeferner:
no need to the code I changed is in this release and covered in the changelog already. Thanks for reviewing

PR #417, opened 2019-12-05
Fixed: InMemory concurrency issues with extended data simultaneous deletes and adds

joeferner:


joeferner:
@mwizeman I removed the unit test for this. It just caused too many concurrency issues and since this really only effects the in-memory graph I'm not all that concerned.

PR #418, opened 2019-12-05
Added: FetchHints.union

joeferner:


core/src/main/java/org/vertexium/FetchHints.java
```diff
@@ -414,4 +417,103 @@ private boolean isEdgeLabelsOfEdgeRefsIncluded(ImmutableSet<String> edgeLabelsOf
...
+                throw new VertexiumException("Incompatible fetch hints to combine (ignoreAdditionalVisibilities).");
+            }
+            ignoreAdditionalVisibilities = fetchHint.isIgnoreAdditionalVisibilities();
+
+            if (includePreviousMetadata != null && includePreviousMetadata != fetchHint.isIncludePreviousMetadata()) {
+                throw new VertexiumException("Incompatible fetch hints to combine (includePreviousMetadata).");
+            }
+            includePreviousMetadata = fetchHint.isIncludePreviousMetadata();
+
+            if (fetchHint.getPropertyNamesToInclude() != null) {
```
mwizeman: Should we also skip this if `isIncludeAllProperties` is true? If the answer to that is yes, should we clean out `propertyNamesToInclude` if we find a true `isIncludeAllProperties` mid-way through the iteration?

core/src/main/java/org/vertexium/FetchHints.java
```diff
@@ -414,4 +417,103 @@ private boolean isEdgeLabelsOfEdgeRefsIncluded(ImmutableSet<String> edgeLabelsOf
...
+
+            if (fetchHint.getMetadataKeysToInclude() != null) {
+                if (metadataKeysToInclude == null) {
+                    metadataKeysToInclude = new HashSet<>(fetchHint.getMetadataKeysToInclude());
+                } else {
+                    metadataKeysToInclude.addAll(fetchHint.getMetadataKeysToInclude());
+                }
+            }
+
+            if (fetchHint.getEdgeLabelsOfEdgeRefsToInclude() != null) {
```
mwizeman: Similar comment as above if we've been asked to include all edges refs.

core/src/main/java/org/vertexium/FetchHints.java
```diff
@@ -414,4 +417,103 @@ private boolean isEdgeLabelsOfEdgeRefsIncluded(ImmutableSet<String> edgeLabelsOf
...
+
+            if (fetchHint.getPropertyNamesToInclude() != null) {
+                if (propertyNamesToInclude == null) {
+                    propertyNamesToInclude = new HashSet<>(fetchHint.getPropertyNamesToInclude());
+                } else {
+                    propertyNamesToInclude.addAll(fetchHint.getPropertyNamesToInclude());
+                }
+            }
+
+            if (fetchHint.getMetadataKeysToInclude() != null) {
```
mwizeman: Similar comment as above if we've been asked to include all metadata.

core/src/main/java/org/vertexium/FetchHints.java
joeferner: I ended up putting this logic in the constructor since it should really work with the builder as well. https://github.com/visallo/vertexium/pull/418/files#diff-804687a9c91cd00ca960efeb91ac898aR73

PR #439, opened 2020-01-15
 Fix Re-Indexing Elements with Hidden Properties

mwizeman:
When re-indexing an element with hidden properties in Elasticsearch, the hidden property field was not being included in the list of fields to add.

### DIFF
--- a/test/src/main/java/org/vertexium/test/GraphTestBase.java
+++ b/test/src/main/java/org/vertexium/test/GraphTestBase.java
@@ -1686,6 +1686,42 @@ public abstract class GraphTestBase {
     }
 
     @Test
+    public void testReindexHiddenProperties() {
+        assumeTrue(isSearchIndexDeleteElementSupported());
+
+        graph.prepareVertex("v1", VISIBILITY_EMPTY)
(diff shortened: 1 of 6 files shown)


test/src/main/java/org/vertexium/test/GraphTestBase.java
```diff
@@ -1685,6 +1685,42 @@ public void testMarkVertexAndPropertiesHidden() {
         assertEquals(30, v1.getPropertyValue("k2", "age"));
     }
 
+    @Test
+    public void testReindexHiddenProperties() {
```
joeferner: Can we move testUpdateVertexWithDeletedElasticsearchDocument to GraphTestBase as well? https://github.com/visallo/vertexium/blob/master/elasticsearch5/search-index/src/test/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndexTest.java#L238

mwizeman: I took a quick look and I think for now those need to stay in ES specific tests due to this line: https://github.com/visallo/vertexium/blob/master/elasticsearch5/search-index/src/test/java/org/vertexium/elasticsearch5/Elasticsearch5SearchIndexTest.java#L239

PR #440, opened 2020-01-16
Added: System property metricRegistryStartConsoleReporter to start the metric registry console reporter automatically

joeferner:


core/src/main/java/org/vertexium/metric/DropWizardMetricRegistry.java
```diff
@@ -25,12 +26,43 @@ public DropWizardMetricRegistry() {
...
+
+        startConsoleReporter(startConsoleReporter);
     }
 
     public MetricRegistry getMetricRegistry() {
         return metricRegistry;
     }
 
+    private void startConsoleReporter(String period) {
+        try {
```
mwizeman: Consider switching this to [support Java duration strings](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-).

PR #445, opened 2020-02-04
Attempt to Automatically Correct GeoPolygon Path Errors

mwizeman:
Improve the GeoPolygon validation to more completely check the GeoJSON rules. Also provide a lenient methods that will attempt to automatically fix some errors in polygons.


core/src/main/java/org/vertexium/type/GeoLine.java
```diff
@@ -40,12 +41,12 @@ public GeoLine(GeoPoint start, GeoPoint end, String description) {
...
     @Override
     public boolean intersects(GeoShape geoShape) {
-        throw new VertexiumException("Not implemented for argument type " + geoShape.getClass().getName());
+        return GeoUtils.intersects(this, geoShape);
     }
 
     @Override
     public boolean within(GeoShape geoShape) {
-        throw new VertexiumException("Not implemented for argument type " + geoShape.getClass().getName());
+        return false;
```
joeferner: Is this true? A line can't be inside a shape?

core/src/main/java/org/vertexium/type/GeoPoint.java
```diff
@@ -109,7 +109,7 @@ public boolean intersects(GeoShape geoShape) {
 
     @Override
     public boolean within(GeoShape geoShape) {
-        throw new VertexiumException("Not implemented for argument type " + geoShape.getClass().getName());
+        return false;
```
joeferner: Same here a point can't be inside a geo poly?

core/src/main/java/org/vertexium/util/GeoUtils.java
```diff
@@ -0,0 +1,144 @@
...
+        if (geometry instanceof GeometryCollection) {
+            GeoCollection geoCollection = new GeoCollection(description);
+            GeometryCollection geometryCollection = (GeometryCollection) geometry;
+            for (int i = 0; i < geometryCollection.getNumGeometries(); i++) {
+                geoCollection.addShape(toGeoShape(geometryCollection.getGeometryN(i), null));
+            }
+            return geoCollection;
+        }
+
+        if (geometry instanceof Polygon) {
```
joeferner: Either make this an `else if` or change the `else if` below to `if`.  I personally like just using `if` since there are return statements inside.

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/KryoVertexiumSerializerTest.java
```diff
@@ -218,23 +218,24 @@ public void before() {
...
-            85, 64, 41, 92, -113, -62, -11, 48, 83, 64, -21, 7, 1, 0, -128, 96,
-            -27, -48, 34, -37, 105, 80, 64, 47, -35, 36, 6, -127, 69, 75, 64, -21,
-            7, 1, 0, -128, -10, 40, 92, -113, -62, 53, 53, 64, 43, -121, 22, -39,
-            -50, 55, 36, 64, 1, 1, 1, 4, -21, 7, 1, 0, -128, -27, -48, 34,
-            -37, -7, 62, 40, 64, -106, 67, -117, 108, -25, 59, 55, 64, -21, 7, 1,
-            0, -128, 92, -113, -62, -11, 40, 44, 65, 64, -18, 124, 63, 53, 94, -70,
-            70, 64, -21, 7, 1, 0, -128, 127, 106, -68, 116, -109, 72, 76, 64, 8,
-            -84, 28, 90, 100, -21, 80, 64, -21, 7, 1, 0, -128, -27, -48, 34, -37,
-            -7, 62, 40, 64, -106, 67, -117, 108, -25, 59, 55, 64, 69, 78, -60
+            115, -12, 1, 2, 1, 1, 1, 4, -9, 7, 1, 0, 0, -128, 125, 63,
```
joeferner: You shouldn't change these bytes here. Create a new set of bytes if you want but these bytes represent what could be in the database and to verify that they will deserialize properly.

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/QuickKryoVertexiumSerializerTest.java
```diff
@@ -349,17 +349,17 @@ public void testCompress() {
...
-                -62, 28, -64, -126, 9, 79, 47, 40, -35, -50, 12, 112, -48, -65, -85, -62,
-                -42, -24, -22, 13, 17, -100, -9, 126, -7, 49, -17, -19, -82, 14, 60, 76,
-                -38, -19, 98, -102, 14, 14, 88, -75, -61, -51, -3, -90, 1, 20, 49, 53,
-                117, 0, -86, -67, 121, -50, 92, 5, -94, -6, -45, 37, -33, 36, -127, 31,
-                -95, 16, -43, 6, -63, 120, 44, 67, -41, 14, 55, 23, -92, -8, -89, -99,
-                -122, -61, 52, -25, -18, -100, -25, -42, -26, 16, -43, 32, -29, 52, 116, 28,
-                29, -34, -43, -40, -101, -58, -19, 114, -125, 8, -42, 103, -19, 41, -103, -20,
-                -31, -29, -64, -79, 70, 38, 42, -27, 117, -128, 3, 86, -19, -82, 126, 71,
-                0, 34, -15, -102, 69
+                18, 43, 125, 50, -117, -65, 48, 50, 49, 50, 50, -78, 124, 103, 103, 100,
```
joeferner: Same here

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/QuickKryoVertexiumSerializerTest.java
```diff
@@ -371,23 +371,24 @@ public void testCompress() {
...
-                -8, 85, 64, 41, 92, -113, -62, -11, 48, 83, 64, -21, 7, 1, 0, -128,
-                96, -27, -48, 34, -37, 105, 80, 64, 47, -35, 36, 6, -127, 69, 75, 64,
-                -21, 7, 1, 0, -128, -10, 40, 92, -113, -62, 53, 53, 64, 43, -121, 22,
-                -39, -50, 55, 36, 64, 1, 1, 1, 4, -21, 7, 1, 0, -128, -27, -48,
-                34, -37, -7, 62, 40, 64, -106, 67, -117, 108, -25, 59, 55, 64, -21, 7,
-                1, 0, -128, 92, -113, -62, -11, 40, 44, 65, 64, -18, 124, 63, 53, 94,
-                -70, 70, 64, -21, 7, 1, 0, -128, 127, 106, -68, 116, -109, 72, 76, 64,
-                8, -84, 28, 90, 100, -21, 80, 64, -21, 7, 1, 0, -128, -27, -48, 34,
-                -37, -7, 62, 40, 64, -106, 67, -117, 108, -25, 59, 55, 64, 69, 78, -60
+                105, 115, -12, 1, 2, 1, 1, 1, 4, -9, 7, 1, 0, 0, -128, 125,
```
joeferner: and here

test/src/test/java/org/vertexium/JavaVertexiumSerializerTest.java
```diff
@@ -387,31 +387,32 @@ public void before() {
...
-            -10, 64, 36, 55, -50, -39, 22, -121, 43, 112, 120, 120, 115, 113, 0, 126,
-            0, 10, 0, 0, 0, 4, 119, 4, 0, 0, 0, 4, 115, 113, 0, 126,
-            0, 13, 112, 64, 40, 62, -7, -37, 34, -48, -27, 64, 55, 59, -25, 108,
-            -117, 67, -106, 112, 115, 113, 0, 126, 0, 13, 112, 64, 65, 44, 40, -11,
-            -62, -113, 92, 64, 70, -70, 94, 53, 63, 124, -18, 112, 115, 113, 0, 126,
-            0, 13, 112, 64, 76, 72, -109, 116, -68, 106, 127, 64, 80, -21, 100, 90,
-            28, -84, 8, 112, 115, 113, 0, 126, 0, 13, 112, 64, 40, 62, -7, -37,
-            34, -48, -27, 64, 55, 59, -25, 108, -117, 67, -106, 112, 120, 116, 0, 3,
-            69, 78, 68
+            0, 4, 68, 0, 8, 108, 97, 116, 105, 116, 117, 100, 101, 68, 0, 9,
```
joeferner: and here

core/src/main/java/org/vertexium/type/GeoLine.java
mwizeman: I was tripped up by this too, but [based on this implementation for `GeoRect`](I was tripped ) I interpret this to be asking if the `GeoShape` is inside a line. If we both thought this felt backwards, maybe we should flip the meaning of this method.

core/src/main/java/org/vertexium/type/GeoPoint.java
mwizeman: Same as above. It feels backwards, but based on the `GeoRect` implementation this appears to be asking if the `GeoShape` is within the point which is always false.

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/KryoVertexiumSerializerTest.java
mwizeman: I think I have to change these bytes. [The polygon being used in the base class is actually invalid](https://github.com/visallo/vertexium/blob/master/test/src/main/java/org/vertexium/test/VertexiumSerializerTestBase.java#L229-L251). The coordinates of the holes are going in the wrong direction and the holes aren't within the outer shell.

core/src/main/java/org/vertexium/util/GeoUtils.java
mwizeman: Good call. Fixed in https://github.com/visallo/vertexium/pull/445/commits/22792556a563f2af4d2573cc86a1469ee575b927

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/KryoVertexiumSerializerTest.java
joeferner: We should handle that case though since all the existing data is going to have it like that. My intent for these byte arrays is that they represent what could exist in real databases.

core/src/main/java/org/vertexium/type/GeoLine.java
joeferner: Yeah maybe we should switch that. Very strange and doesn't read very well.

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/KryoVertexiumSerializerTest.java
mwizeman: I was worried about that too. The existing data deserializes just fine with these changes. The [issue is here where the test tries to build a polygon](https://github.com/visallo/vertexium/blob/master/test/src/main/java/org/vertexium/test/VertexiumSerializerTestBase.java#L229-L251) to compare the bytes to. In order to verify, I fixed that polygon but left the bytes the same. Deserialization worked and [the test failed here](https://github.com/visallo/vertexium/blob/master/test/src/main/java/org/vertexium/test/VertexiumSerializerTestBase.java#L261) trying to compare the coordinates.

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/KryoVertexiumSerializerTest.java
```diff
@@ -218,24 +244,23 @@ public void before() {
...
-            -9, 7, 1, 0, 0, -128, 84, -29, -91, -101, -60, 64, 57, 64, -86, -15,
-            -46, 77, 98, 64, 73, 64, -9, 7, 1, 0, 0, -128, -6, 126, 106, -68,
-            116, 51, 52, 64, -80, 114, 104, -111, -19, -84, 65, 64, 1, 1, 1, 4,
-            -9, 7, 1, 0, 0, -128, -27, -48, 34, -37, -7, 62, 40, 64, -106, 67,
-            -117, 108, -25, 59, 55, 64, -9, 7, 1, 0, 0, -128, 92, -113, -62, -11,
-            40, 44, 65, 64, 119, -66, -97, 26, 47, 29, 84, 64, -9, 7, 1, 0,
-            0, -128, 127, 106, -68, 116, -109, 72, 76, 64, 8, -84, 28, 90, 100, -21,
-            80, 64, -9, 7, 1, 0, 0, -128, -27, -48, 34, -37, -7, 62, 40, 64,
-            -106, 67, -117, 108, -25, 59, 55, 64, 69, 78, -60
+            115, -12, 1, 2, 1, 1, 1, 4, -21, 7, 1, 0, -128, -47, 34, -37,
```
joeferner: Still need changes?

core/src/main/java/org/vertexium/type/GeoLine.java
mwizeman: Done

core/src/main/java/org/vertexium/type/GeoPoint.java
mwizeman: Done

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/KryoVertexiumSerializerTest.java
mwizeman: Fixed. I have new legacy methods that test deserializing the exact same bytes to ensure it still works. In addition, I added a new method in GraphTestBase that tests mutating and re-indexing a legacy vertex with invalid geometry.

kryo-serializer/src/test/java/org/vertexium/serializer/kryo/KryoVertexiumSerializerTest.java
mwizeman: Nope. I've got it worked out properly now after a few changes.

PR #448, opened 2020-02-12
Add StringStreamingPropertyValue for Faster Indexing

mwizeman:
Add a specialized version on StreamingPropertyValue to be more performant in the case where the value is a String already in memory


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.9.r43
```
joeferner: what version?

mwizeman: Oops. `vi` typo.

PR #450, opened 2020-02-21
Improve Detection of Elasticsearch Scroll Leaks

mwizeman:
Improve detection and logging of Elasticsearch scrolls that are not closed by Vertexium clients. Fixed a few leaks present in the test code and also added a check after every test to fail if a leak is introduced.

### DIFF
--- a/elasticsearch5/search-index/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
+++ b/elasticsearch5/search-index/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
@@ -8,6 +8,7 @@ import com.vividsolutions.jts.geom.Coordinate;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.standard.StandardAnalyzer;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+import org.elasticsearch.action.search.ClearScrollResponse;
 import org.elasticsearch.action.search.SearchRequestBuilder;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.client.Client;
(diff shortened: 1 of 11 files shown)


elasticsearch5/search-index/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
```diff
@@ -480,9 +481,12 @@ protected IdStrategy getIdStrategy() {
 
     private void closeScroll(String scrollId) {
         try {
-            client.prepareClearScroll()
+            ClearScrollResponse clearScrollResponse = client.prepareClearScroll()
                 .addScrollId(scrollId)
                 .execute().actionGet();
+            if (!clearScrollResponse.isSucceeded()) {
+                LOGGER.warn("Unable to clear scroll " + scrollId);
```
joeferner: any possibility of logging the response here

elasticsearch7/search-index/src/main/java/org/vertexium/elasticsearch7/ElasticsearchSearchQueryBase.java
```diff
@@ -484,9 +485,12 @@ protected IdStrategy getIdStrategy() {
 
     private void closeScroll(String scrollId) {
         try {
-            client.prepareClearScroll()
+            ClearScrollResponse clearScrollResponse = client.prepareClearScroll()
                 .addScrollId(scrollId)
                 .execute().actionGet();
+            if (!clearScrollResponse.isSucceeded()) {
+                LOGGER.warn("Unable to clear scroll " + scrollId);
```
joeferner: log response?

elasticsearch5/search-index/src/main/java/org/vertexium/elasticsearch5/ElasticsearchSearchQueryBase.java
mwizeman: Looking at the `ClearScrollResponse`, it appears to only have two fields: `succeeded` and `num_freed`. I don't think we can provide any other help in that log message

PR #452, opened 2020-03-09
Fix renaming of elementing during search update

dspoja:
This PR fixes an issue with updating a fvisibility on a field and then updating it again. The result was that the updated field would get its visiblity set correctly, but the value would be set to `null` even though the value is not null. This fix ensures that does not happen.


sfeng88:
@dspoja Can you take a look at the test failure please?

dspoja:
The test failures were a result of Travis running the some of the tests in parallel and ES tests getting confused with which ES instance to use, so reruning the build resolved the test failure.

PR #459, opened 2020-03-13
Added: Add fetch hint to exclude edge vertex ids or edge ids from edge references

joeferner:


core/src/main/java/org/vertexium/FetchHintsBuilder.java
```diff
@@ -44,6 +48,9 @@ public FetchHints build() {
         if (!isIncludeProperties() && isIncludePropertyMetadata()) {
             includeAllProperties = true;
         }
+        if (!includeEdgeIds && !includeEdgeVertexIds) {
```
mwizeman: Technically not an error if they also excluded edge refs.

joeferner: good point. I'll make that change

PR #463, opened 2020-03-20
Edge label index

joeferner:
The gist of this PR is to not store the same label string over and over again and to reduce the number of round trips from string to byte[]. I started this with a simple string intern here https://github.com/visallo/vertexium/blob/master/accumulo/iterators/src/main/java/org/vertexium/accumulo/iterator/model/EdgeInfo.java#L34 but ended up having to keep the byte[] (https://github.com/visallo/vertexium/blob/master/accumulo/iterators/src/main/java/org/vertexium/accumulo/iterator/model/EdgeInfo.java#L12) around which was actually causing most of the memory usage. With this PR it removes the need to keep the byte array around and reduces the number of label string copies to one per unique label.


accumulo/iterators/src/main/java/org/vertexium/accumulo/iterator/model/EdgeLabels.java
```diff
@@ -0,0 +1,91 @@
...
+        labels.add(Arrays.copyOfRange(bytes, offset, offset + length));
+        return result;
+    }
+
+    public boolean contains(byte[] bytes) {
+        return indexOf(bytes) >= 0;
+    }
+
+    public int indexOf(byte[] bytes) {
+        int i = 0;
```
mwizeman: This method should call the one below with offset 0 and length of `bytes.length`

